PR #695, opened 2023-12-13
fix number unit padding

RubirajAccenture:


packages/ui-library/src/components/forms/number-input/index.css.ts
```diff
@@ -1,181 +1,173 @@
...
         }
       }
+    }
 
-      &:focus-within {
-        outline: ${Input.Default.Focus.width} ${Input.Default.Focus.style} ${Input.Default.Focus.color};
-        background-color: ${SurfaceFill.Default.Focus};
+    .unit,
+    input {
+      &.sm {
```
davidken91, line 83: We may want to consider making the width of the input 100% for all sizes. I noticed that we can't click (or focus) on the input for most of the space it fills because the width is small. It may look ok visually, but it might be a better user experience if they have more space to click on the input. Hopefully the screenshot below helps clarify what I mean
Screenshot 2024-01-22 at 14 12 02

davidken91: Making this change will also fix the readonly version:
Before:
Screenshot 2024-01-22 at 14 16 15
After:
Screenshot 2024-01-22 at 14 16 20

davidken91 (commented):
I think the input field of 'no unit' variant needs to match the left padding of 'unit suffix'
Screenshot 2024-01-22 at 14 21 36

packages/ui-library/src/components/forms/number-input/index.css.ts
ChristianHoffmannS2: To the first topic: if the input becomes full width, we can not center it in split mode like seen in figma. its actually the opposed thing we need here.... it needs to shrink/grow depending on its content. the entire wrapper would become the right cursors, and no matter where you click (except buttons) it would gain focus.

This was discussed to be done in a follow up ticket. 

but the read only paddings needs to be adressed

JpunktWpunkt (dismissed):

PR #703, opened 2023-12-14
feat(ui-library): add figma plugin to storybook

kaikdi:


davidken91 (commented):
Looks good. I'm guessing that we don't link to any design file for the Slider components because it's still a work in progress?

kaikdi:
> Looks good. I'm guessing that we don't link to any design file for the Slider components because it's still a work in progress?

yes, I couldn't find some designs, but we can add and adjust anytime

davidken91 (approved):

PR #716, opened 2023-12-21
feat(storybook): add tooltip docs

bar-tay:


packages/ui-library/src/components/feedback/tooltip/index.stories.ts
```diff
@@ -1,123 +1,306 @@
...
+ * ### Placement
+ * The Tooltip component can be positioned in any of the following placements: top, top-start, top-end, right, right-start, right-end, bottom, bottom-start, bottom-end, left, left-start, and left-end.
+ */
+
+export const Placement = () => {
+  return html`
+    ${sharedStyles}
+
+    <div class="wrapper">
+      ${Tooltip({
```
davidken91, line 138: I hope you don't mind, but I updated this PR with the latest from `develop`. I think the only recommendation I would have here would be to maybe add a label to each of these examples. On first glance it can be hard to tell which blue box corresponds to which tooltip placement. So maybe add 'a Top Placement' before we call the Tooltip function in side "wrapper" and do some basic styling to make it clear what each example does

bar-tay: Yes, good point. I will implement that in my next commit

thrbnhrtmnn (changes requested):
Left some comments in the ticket.

Also the tooltip-bubble documentation is part of #715 , which Rubi was already starting. Most of the feedback in the tooltip issue also applies to the tooltip bubble. Maybe you should align quickly with @RubirajAccenture

bar-tay:
> Left some comments in the ticket.
> 
> Also the tooltip-bubble documentation is part of #715 , which Rubi was already starting. Most of the feedback in the tooltip issue also applies to the tooltip bubble. Maybe you should align quickly with @RubirajAccenture

Okay. Will do

thrbnhrtmnn (dismissed):

thrbnhrtmnn (approved):

JpunktWpunkt (approved):

PR #718, opened 2023-12-22
fixing icon compile rimraf, fixing missing source icons and add lazy loading to icons and run test against build

ChristianHoffmannS2:
This works great on storybook and js-example-app and reduces the buildsize of the initial js file which gets loaded

~~But sadly, we found out that our tests are build differently on the fly with esbuild.
In order to get this working, we have to rewire tests to use our build instead and rebase this branch onto that~~
(Done: works)


packages/ui-library/src/components/feedback/loader/index.ts
```diff
@@ -4,9 +4,8 @@ import { customElement, property } from 'lit/decorators.js';
 import { styleCustomLight, styleCustomDark } from './index.css';
 import { ActionSizesType, FeedbackVariantType } from '../../../globals/types';
 import { ThemeType } from '../../../foundation/_tokens-generated/index.themes';
-import { genericBlrComponentRenderer } from '../../../utils/typesafe-generic-component-renderer';
 
-const TAG_NAME = 'blr-loader';
+export const TAG_NAME = 'blr-loader';
```
kaikdi, line 8: maybe we should rename these because now we have multiple `TAG_NAME`s in our global namespace

ChristianHoffmannS2: we should rename it: yes.

does it collide or pollute? no :) 

the index.ts defines whats exported to build, and its not imported/exported there.
also the build is setup up to not use global name space.

ChristianHoffmannS2: double checked: no TAG_NAME inside the build at all :)

kaikdi: To be honest I didn't have the build in mind, I just meant that if you try to import `TAG_NAME` you now the `ts-server` suggestions are confusing

ChristianHoffmannS2: yes, i completely agree on that point. i also dont want to know what happens if you move component folders around. they should be unique. but maybe not on this branch wich is already hard to rebase on

PR #719, opened 2023-12-25
NumberInput - Fix the disappearing stepper buttons and hiding placeholder text

usmanreal1:
This MR fixes two issues:

- The disappearing stepper button upon two actions:
   - Setting the unit to `undefined`
   - Setting `readonly` to `true`

It extracts the stepper rendering method out of the conditional rendering of the unit text.
- Placeholder unable to appear because the default value is not undefined

It adds a condition to render the placeholder text for 0 value.


ChristianHoffmannS2 (approved):

ChristianHoffmannS2:
Thanks for contribution :)

JpunktWpunkt (approved):
lgtm :)

PR #756, opened 2024-01-10
feat(storybook): add icon docs

bar-tay:


JpunktWpunkt (dismissed):
lgtm 🚀

thrbnhrtmnn (changes requested):
Please move the ignoreSize prop to Appearance (under size) for the Icon documentation

thrbnhrtmnn (changes requested):
Now I saw that the default value for the size prop is wrong. It should be md and not lg. Sorry, should have seen that before.

thrbnhrtmnn (dismissed):

packages/ui-library/src/components/ui/icon/index.ts
```diff
@@ -8,15 +8,18 @@ import { ClassMapDirective } from 'lit-html/directives/class-map';
...
+import { ThemeType } from '../../../foundation/_tokens-generated/index.themes';
 
 @customElement(TAG_NAME)
 export class BlrIcon extends LitElement {
   static styles = [styleCustom];
 
   @property() icon: IconType = 'blr360Xs';
   @property() size: SizesType = 'md';
-
+  @property() arialabel?: string;
```
ChristianHoffmannS2, line 19: we intentionally removed this one, because we figured out a way how to add html attributes which are not component properties the right way. so we will not need it as property anymore

packages/ui-library/src/components/ui/icon/index.ts
```diff
@@ -8,15 +8,18 @@ import { ClassMapDirective } from 'lit-html/directives/class-map';
...
   @property() ignoreSize?: boolean = false;
+  @property() onClick?: () => void;
```
ChristianHoffmannS2, line 21: not sure if this is up to date.
on develop this event is called blrClick
https://github.com/deven-org/B01LER-Kitchen/blob/develop/packages/ui-library/src/components/ui/icon/index.ts

bar-tay: done

packages/ui-library/src/components/ui/icon/index.ts
bar-tay: done

bar-tay: So we also dont need it reperesentet anyomre in the documentation file as well?

packages/ui-library/src/components/ui/icon/index.ts
bar-tay: done

packages/ui-library/src/components/ui/icon/index.ts
ChristianHoffmannS2: yes :)

thrbnhrtmnn (approved):

ChristianHoffmannS2 (approved):

PR #768, opened 2024-01-11
Remove internal caption tokens (Figma only)

larserbach:
I took a look, and the caption tokens were already removed. Only the $themes.json was still holding some information on these, which get removed with this PR.
It is important to note: Ususally we have conflicts on themes.json if a token changed. We should always pick:
- development if the branch was created before this PR was merged
- always pick incoming branch if the branch was created after this PR

Hope this makes sense


larserbach:
@ChristianHoffmannS2  Lit-Analyze / analyze (pull_request) is failing. I need your help to resolve the issue.

ChristianHoffmannS2 (approved):

PR #780, opened 2024-01-12
Select: html options instead of array

ChristianHoffmannS2:
Still needs some attention to tests and stories, but works already


packages/js-example-app/package.json
```diff
@@ -14,5 +14,10 @@
   "license": "ISC",
   "dependencies": {
     "@boiler/ui-library": "0.0.1"
+  },
+  "devDependencies": {
+    "http-server": "^14.1.1",
+    "webpack": "^5.89.0",
+    "webpack-dev-server": "^4.15.1"
```
ChristianHoffmannS2, line 21: not sure why that, but the plan was to have the same webpack version accross all example apps. same for the dev server, hence to have them in the global package.json

ChristianHoffmannS2: i think if you reset that file, the build will work again in the CI.

its complaining because after install in CI, the yarn.lock will be altered. it expects that commits not only include the dep. changes, but also the matching yarn.lock to those changes

davidken91: Ok, I'll remove it but the js example app can now no longer find webpack or the http server

ChristianHoffmannS2: lets have a look into that together :)

davidken91: I've actually fixed it. I was making a stupid mistake by running the command in that directory instead of the root directory  
🙄. I just need to make sure that the new click event works in the example app and then it's finished

JpunktWpunkt (approved):

davidken91 (approved):

Issue #419, opened 2023-10-06
Storybook Documentation - Component: Toggle Switch


angelicahoyss:
### Description / User Story

With this task we want to make the Toggle Switch Component Docs page in Storybook look like the [template](https://www.figma.com/file/Sg2Uox6PchuAsMKY4ShXku/StorybookTemplate?type=design&node-id=1-1072&mode=design&t=LeRpJsk2yo3q8jHL-0) from design.

---

### Acceptance Criteria

- [x] Component Page has the same structure as the design template (in the following order):
    - [x]  Headline & Intro Text incl. a list of anchor links to all following sections
    - [x]  Docs Area with Example and all Props (see which props should be displayed for this component in the section below)
    - [x]  Appearance
      - [x] SizeVariant (incl. sizes = SM, MD, LG)
      - [ ] ControlPosition
        - Control: Left
        - Control: Right
    - [x] Content / Settings
      - [x]  Active
      - [x]  Inactive
      - [x] OnLabel
      - [x] OffLabel
    - [x]  States
        - [x] Disabled  (incl. disabled = true)
        - [ ] Readonly  (incl. readonly = true)
    - [x] Dependencies
        - [ ]  FormLabel
          - A Toggle Switch with a label
          - A Toggle Switch with no label
        - [x]  FormCaption
            - A Toggle Switch with a hint
- [x]  Component Page includes the texts as defined in the section below
- [x]  Component Page text styles (uppercase, lowercase, spacing) correspond the the design template
- [x]  Component Page spacing corresponds to design template
- [x]  Component Page colors correspond to design template

---

### Props
See [this table](https://ts.accenture.com/:x:/r/sites/b01ler-core/_layouts/15/Doc.aspx?sourcedoc=%7Bb2cc9fb8-5c39-4049-9652-a4e6bbe09ef1%7D&action=editnew) page "Checkbox" for all props that should be displayed for the component, incl. their description, default values and the type of controls.

If the props in the list are not the same as the component currently uses, just leave a comment in column G in the excel and a follow-up task will be created. Fixing the components is not part of this ticket.

---

### Content / Texts

[...]

PR #783, opened 2024-01-15
Feature/419 storybook toggle switch

JpunktWpunkt:
#419


thrbnhrtmnn (dismissed):

packages/ui-library/src/components/forms/toggle-switch/index.stories.ts
```diff
@@ -4,76 +4,463 @@ import { BlrToggleSwitchRenderFunction } from './renderFunction';
...
   },
   parameters: {
     design: {
       type: 'figma',
       url: 'https://www.figma.com/file/C4vgEKz8mKyulJ4gm3Qdql/%F0%9F%AB%A7-%5BBLR%5D-The-B01LER?node-id=3618%3A125201&mode=dev',
     },
+    layout: 'centered',
+    docs: {
+      description: {
+        component: `<Markdown>
```
bar-tay, line 185:
```suggestion
        component: `<markdown>
```

bar-tay: Pls change to lower case since this causes an error that can only be seen wehen opening the dev tools. I had the same review suggestion before in one my of my tickets.

packages/ui-library/src/components/forms/toggle-switch/index.stories.ts
```diff
@@ -4,76 +4,463 @@ import { BlrToggleSwitchRenderFunction } from './renderFunction';
...
+        `,
+      },
+    },
   },
 };
 
 export const BlrToggleSwitch = (params: BlrToggleSwitchType) => BlrToggleSwitchRenderFunction(params);
 
 BlrToggleSwitch.storyName = 'ToggleSwitch';
+BlrToggleSwitch.args = defaultParams;
```
bar-tay, line 210:
```suggestion
export const ToggleSwitch = (params: BlrToggleSwitchType) => BlrToggleSwitchRenderFunction(params);
ToggleSwitch.args = defaultParams;
```

bar-tay: We can just use the more understandable naming version as the component declaration.

bar-tay (changes requested):

packages/ui-library/src/components/forms/toggle-switch/index.stories.ts
JpunktWpunkt: I saw that. I'm not sure if that solve that problem, because now we got another warning in the dev tools.

bar-tay: Yes, the other error persists but I talked too Christian about it and its still better then having both errors

packages/ui-library/src/components/forms/toggle-switch/index.stories.ts
JpunktWpunkt: done

packages/ui-library/src/components/forms/toggle-switch/index.stories.ts
JpunktWpunkt: done

bar-tay (approved):

thrbnhrtmnn (approved):

PR #787, opened 2024-01-15
icon uses correct event handling for blrClick - fixes password

ChristianHoffmannS2:
text-Input password functionality was broken


JpunktWpunkt (commented):
great thing, the only thing i noticed is that when the focus is active, you have to click twice on the icon for anything to happen

ChristianHoffmannS2:
> great thing, the only thing i noticed is that when the focus is active, you have to click twice on the icon for anything to happen

true, but i wouldt not fix it here. my aim was not to fix text-input but rather finish the icon refactoring :)

seemann (approved):

davidken91 (approved):

PR #814, opened 2024-01-19
Code snipped fix

ChristianHoffmannS2:
this pr changes the render function to only use dots in two cases:

if the value is an object/array or if the value is a boolean false. 
In both cases, you will not see key/value in storybooks code preview.

Since we want to move away from arrays anyway, this is actually not a problem for us


JpunktWpunkt (approved):
n1ce

davidken91 (approved):

PR #830, opened 2024-01-24
Fix/597 form label padding

davidken91:


packages/ui-library/src/components/forms/number-input/index.css.ts
```diff
@@ -7,6 +7,27 @@ export const { tokenizedLight: wrapperLight, tokenizedDark: wrapperDark } = rend
   const { NumberInput } = componentTokens.Forms;
 
   return typeSafeNestedCss`
+  .blr-number-input {
+    &.sm {
+      > blr-form-label {
```
ChristianHoffmannS2, line 12: we always should start these with &

& > blr-form-label

and also: components should not overwrite each others internals. can we have that token set in the blr-form-label? it would be much less code as well

davidken91: Screenshot 2024-01-24 at 17 11 51
Yeah I had the same thoughts, but I was asked to do it on a per component basis. I guess the use will have more granular control over it in the future

PR #855, opened 2024-01-31
Fix/597 form label padding

davidken91:


bar-tay (approved):
lgtm :)

packages/ui-library/src/components/forms/number-input/index.ts
```diff
@@ -229,37 +229,41 @@ export class BlrNumberInput extends LitElement {
...
+                  : nothing}
+                step="${this.step || nothing}"
+                ?disabled="${this.disabled}"
+                ?readonly="${this.readonly}"
+                ?required="${this.required}"
+                hasError="${this.hasError}"
+                @change=${this.handleChange}
+                placeholder=${this.placeholder || nothing}
+              />
+              ${this.unit !== undefined && this.unit.length
```
angsherpa456, line 261: hasUnit is already there for this check.

packages/ui-library/src/components/forms/text-input/index.ts
```diff
@@ -131,18 +131,20 @@ export class BlrTextInput extends LitElement {
...
-                  labelAppendix: this.labelAppendix,
-                  forValue: this.textInputId,
-                  theme: this.theme,
-                  variant: this.hasError ? 'error' : 'label',
-                })}
-              `
-            : nothing}
+          <div class="label-wrapper">
+            ${this.hasLabel
+              ? html`
```
angsherpa456, line 136: is html thingy required here because in this file 'packages/ui-library/src/components/forms/select/index.ts' in line number 161, same logic exist but with the html part.

packages/ui-library/src/components/forms/textarea/index.ts
```diff
@@ -154,14 +154,16 @@ export class BlrTextarea extends LitElement {
...
-          ${BlrFormLabelRenderFunction({
-            labelText: this.label,
-            labelSize: this.size,
-            labelAppendix: this.labelAppendix,
-            forValue: this.textareaId,
-            theme: this.theme,
-            variant: this.hasError ? 'error' : 'label',
-          })}
+          <div class="label-wrapper">
+            ${BlrFormLabelRenderFunction({
```
angsherpa456, line 158: Is this intentional that label presence is not checked for textarea but is checked for text-input? Right now the hasLabel property is a dead code as it is not used anywhere in this file except during declaration.

davidken91: Ok, I see what you mean. I'll add in a check for hasLabel and include label-wrapper within that check. I can apply this to all components that use it

packages/ui-library/src/components/forms/textarea/index.ts
```diff
@@ -154,16 +154,23 @@ export class BlrTextarea extends LitElement {
...
-          </div>
+        ${
+          this.hasLabel
+            ? html`
+                <div class="label-wrapper">
+                  ${BlrFormLabelRenderFunction({
+                    labelText: this.label,
+                    labelSize: this.size,
+                    labelAppendix: this.labelAppendix,
+                    forValue: this.textareaId,
```
angsherpa456, line 165: Can textareaId be written as camelCase, just like textInputId?

packages/ui-library/src/components/forms/textarea/index.ts
```diff
@@ -187,18 +194,22 @@ ${this.value}
...
-                })
-              : nothing}
+            ${
+              this.hasHint || this.hasError
+                ? BlrFormCaptionGroupRenderFunction({ size: this.size }, captionContent)
+                : nothing
+            }
+            ${
+              this.showCounter
+                ? BlrCounterRenderFunction({
```
angsherpa456, line 204: In other places for the renderFunction the 'html' is used, example in this file packages/ui-library/src/components/forms/textarea/index.ts in line 159 but here for this renderFunction 'html' thingy is not there.  I think if 'html' is not required then it's better to be removed.

JpunktWpunkt (approved):

packages/ui-library/src/components/forms/textarea/index.ts
davidken91: Ok I can change it, but isn't the textarea element one word? I know that in storybook we use Text Area to label the component but I'm not sure if that's correct

packages/ui-library/src/components/forms/textarea/index.ts
angsherpa456: Ok, thanks for the clarification.

packages/ui-library/src/components/forms/textarea/index.ts
angsherpa456: Hmm, yes, seems like it's written like what we have now, so the current style should be fine.

bar-tay (approved):

angsherpa456 (approved):
LGTM

packages/ui-library/src/components/forms/textarea/index.ts
JpunktWpunkt: Yes, this is exactly how we have treated it in the past, as one word. I remember that I asking the same thing once.

Issue #798, opened 2024-01-17
Form Caption - icon in sm is not center aligned


thrbnhrtmnn:
#### Description / User Story

The icon in the Form Caption SM seems to be a bit off in code when compared to design. It seems like the icon is not center aligned, but bottom right aligned in the wrapping 14x14 px container. The issue might also be there in other sizes, but not that recognizable.

----
#### Acceptance Criteria
- [ ] The icon placement within its container within the Form Caption has been investigated
- [ ] The icon has been adjusted so it is perfectly aligned with design
- [ ] Property name "ignoreSize" has been changed to "fillParent" for all components
- [ ] "fillParent" default should be true for icon component
- [ ] "sizeVariant" prop should only be displayed in prop panel of icon component in Storybook, if fillParent=false 

----
#### Background information
- _Add relevant information here_
- ...

larserbach:
I took a look. (on Render with Firefox) The vertical alignment is correct in all size variants.
The technical solution differs from the design solution. There is no wrapper for the icon created, the padding is applied on the icon directly. Yet the visual result is the same as in design, therefor the ticket can be closed from my perspective.

PR #871, opened 2024-02-03
fix(ui-library): align icon center in sm (#798)

angsherpa456:


packages/ui-library/src/components/internal-components/form-caption/index.css.ts
```diff
@@ -20,7 +20,6 @@ export const { tokenizedLight: captionLight, tokenizedDark: captionDark } = rend
         gap: ${CaptionComponent.Container.ItemSpacing.SM};
 
         .blr-icon {
-          padding-top: ${CaptionComponent.IconWrapper.PaddingTop.SM};
```
ChristianHoffmannS2, line 23: @larserbach opinion?

larserbach: The removal of this line seems wrong.
The intention of this padding-top is to align the icon against the first line of text.
But it might be important to understand, that by design, the icon is wrapped inside an "iconWrapper". The padding is assigned on "iconWrapper" not the icon itself. (I believe we should not mess with the css of included components if possible).
image

bar-tay (approved):

angsherpa456:
Resolved in another ticket

PR #881, opened 2024-02-05
fix(storybook): adjust component size

### LINKED ISSUE #576: Loader - align padding in code and design
#### Description / User Story

Currently the padding in design and code are applied differently. We need to align the way the padding gets applied.

----
#### Acceptance Criteria
- [ ] The loader has the same padding as in design (in design the padding around the loader elements is 1px in SM, 2px in MD and 3px in LG, but in code there is no padding)
- [ ] The total size of the loader did not change and is still (16x16 in SM, 24x24 in MD and 32x32 in LG)
- [ ] In case props were changed: [Props Excel](https://ts.accenture.com/:x:/r/sites/b01ler-core/Shared%20Documents/Documentation_Props.xlsx?d=wb2cc9fb85c3940499652a4e6bbe09ef1&csf=1&web=1&e=bhRYao) has been updated / comments have been resolved and props changed from red to black font color > @thrbnhrtmnn or @angelicahoyss can support here

----
#### Background information
- This was found after #549  were done, therefor the errors are likely caused by semantic tokens and not by component tokens
- Token name example for LG: --feedback-loader-container-padding-lg, 3px

bar-tay:
Closes #576


ChristianHoffmannS2 (approved):

packages/ui-library/src/components/feedback/loader/index.css.ts
thrbnhrtmnn, line 1: Hey @bar-tay ,

it looks good, but if I inspect the size for SM and LG Loader, the values are a bit off, when compared to design.

**SM Loader:** 
sm_loader
- the padding is 1.008 px instead of 1
- the div is 13.969x13.969px instead of 14x14px

**LG Loader:**
lg_loader
- padding is 3.008px instead of 3px
- the div is 25.969x25.969px instead of 26x26px

Is there a way to fix this or is this something we have to live with?

bar-tay: Yeah, I had a talk with Lars and it is something we need to live with since we are getting relative values from Figma.

Issue #482, opened 2023-10-19
TextInput - add missing events


thrbnhrtmnn:
#### Description / User Story

The TextInput should at least have the following events:
- onChange
- onFocus
- onSelect
- onBlur

The events should be implemented similar to how it is currently already done for Checkbox, TextButton and IconButton. You can also see how the events are working in the vanilla js example application.

One finding we had there was that the events can not have the same name as the html events.

----
#### Acceptance Criteria
- [ ] onChange event was implemented
- [ ] onFocus event was implemented
- [ ] onSelect event was implemented
- [ ] onBlut event was implemented
- [ ] [Props Excel](https://ts.accenture.com/:x:/r/sites/b01ler-core/Shared%20Documents/Documentation_Props.xlsx?d=wb2cc9fb85c3940499652a4e6bbe09ef1&csf=1&web=1&e=bhRYao) has been updated / comments have been resolved and props changed from red to black font color > @thrbnhrtmnn or @angelicahoyss cab support here

----
#### Background information
- [Link](https://ts.accenture.com/:x:/r/sites/b01ler-core/Shared%20Documents/Documentation_Props.xlsx?d=wb2cc9fb85c3940499652a4e6bbe09ef1&csf=1&web=1&e=ksr0Vb) to props documentation (consider All Props page & Component page)

PR #890, opened 2024-02-07
fix(ui): add eventListeners to Textinput (#482)

angsherpa456:


davidken91 (commented):
I just tested this out in the example js app and it works as expected 👍🏻 I can share with you the code that I used to integrate it or I can even add an additional commit to this branch if you like. The only thing I'm not sure about is the use of "@select", it doesn't seem to do anything in the example app. I think that this use case will be covered by the 'focus' event

angsherpa456:
> anything in the example app. I think that this use case will be covered by the 'f

Thanks, please feel free to add the additional commit. Regarding the select, personally I also did not see any practical use case for select but since it was there in the acceptance criteria, I added there. It can be removed, like you said focus event should be pretty much enough.

bar-tay (approved):

PR #893, opened 2024-02-08
Fix/form caption accessibility

davidken91:


packages/ui-library/src/components/internal-components/form-caption/index.stories.ts
```diff
@@ -58,14 +58,6 @@ export default {
         category: 'Content / Setting',
       },
     },
-    arialabel: {
```
angsherpa456, line 61: Does this mean that this property needs to be removed from all other components or just the formCaption?

RubirajAccenture (approved):

JpunktWpunkt (approved):

PR #895, opened 2024-02-08
fix(ui): adjust naming in documentation of Textarea (#485) (#885)

### LINKED ISSUE #485: TextArea - add missing events
#### Description / User Story

The TextArea should at least have the following events:
- ~onChange~ > blrChange
- ~onFocus~ > blrFocus
- ~onSelect~ > blrSelect
- ~onBlur~ > blrBlur

The events should be implemented similar to how it is currently already done for Checkbox, TextButton and IconButton. You can also see how the events are working in the vanilla js example application.

One finding we had there was that the events can not have the same name as the html events.

----
#### Acceptance Criteria
- [x] blrChange event was implemented
- [x] blrFocus event was implemented
- [x] blrSelect event was implemented
- [x] blrBlut event was implemented
- [ ] [Props Excel](https://ts.accenture.com/:x:/r/sites/b01ler-core/Shared%20Documents/Documentation_Props.xlsx?d=wb2cc9fb85c3940499652a4e6bbe09ef1&csf=1&web=1&e=bhRYao) has been updated / comments have been resolved and props changed from red to black font color > @thrbnhrtmnn or @angelicahoyss cab support here

----
#### Background information
- [Link](https://ts.accenture.com/:x:/r/sites/b01ler-core/Shared%20Documents/Documentation_Props.xlsx?d=wb2cc9fb85c3940499652a4e6bbe09ef1&csf=1&web=1&e=ksr0Vb) to props documentation (consider All Props page & Component page)

angsherpa456:
closes #485

### DIFF
--- a/packages/ui-library/src/components/textarea/index.stories.ts
+++ b/packages/ui-library/src/components/textarea/index.stories.ts
@@ -319,34 +319,30 @@ export default {
     },
 
     // Events
-    onChange: {
-      name: 'onChange',
+    blrChange: {
       description: 'Fires when the value changes.',
-      action: 'onChange',
+      action: 'blrChange',
       table: {
         category: 'Events',
       },
     },
-    onFocus: {
-      name: 'onFocus',
-      description: 'Fires when the component is focused.',
-      action: 'onFocus',
-      table: {
-        category: 'Events',
-      },
-    },
-    onBlur: {
-      name: 'onBlur',
-      description: 'Fires when the component lost focus.',
-      action: 'onBlur',
-      table: {
-        category: 'Events',
-      },
-    },
-    onSelect: {
-      name: 'onSelect',
+    blrSelect: {
       description: 'Fires when some text is selected.',
-      action: 'onSelect',
+      action: 'blrSelect',
+      table: {
+        category: 'Events',
+      },
+    },
+    blrFocus: {
+      description: 'Fires when the component is focused.',
+      action: 'blrFocus',
+      table: {
+        category: 'Events',
+      },
+    },
+    blrBlur: {
+      description: 'Fires when the component lost focus.',
+      action: 'blrBlur',
       table: {
         category: 'Events',
       },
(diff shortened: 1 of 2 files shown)


davidken91:
This looks fine to me, maybe just add the events for textarea to the example js app like the way we do it for text input

packages/ui-library/src/components/textarea/index.stories.ts
```diff
@@ -319,34 +319,34 @@ export default {
     },
 
     // Events
-    onChange: {
-      name: 'onChange',
+    blrChange: {
+      name: 'blrChange',
```
JpunktWpunkt, line 323: name overwrite the origin property name and show us in storybook what we want to see (instead of the origin).  I used it sometimes to show sizeVariant in the documentation correctly. In that case we don't need this line because the both have the same value."name"  is no longer needed and can be removed.

packages/ui-library/src/components/textarea/index.stories.ts
```diff
@@ -319,34 +319,34 @@ export default {
...
       table: {
         category: 'Events',
       },
     },
-    onFocus: {
-      name: 'onFocus',
-      description: 'Fires when the component is focused.',
-      action: 'onFocus',
+    blrSelect: {
+      name: 'blrSelect',
```
JpunktWpunkt, line 331: name overwrite the origin property name and show us in storybook what we want to see (instead of the origin).  I used it sometimes to show sizeVariant in the documentation correctly. In that case we don't need this line because the both have the same value."name"  is no longer needed and can be removed.

packages/ui-library/src/components/textarea/index.stories.ts
```diff
@@ -319,34 +319,34 @@ export default {
...
       table: {
         category: 'Events',
       },
     },
-    onSelect: {
-      name: 'onSelect',
-      description: 'Fires when some text is selected.',
-      action: 'onSelect',
+    blrBlur: {
+      name: 'blrBlur',
```
JpunktWpunkt, line 347: name overwrite the origin property name and show us in storybook what we want to see (instead of the origin).  I used it sometimes to show sizeVariant in the documentation correctly. In that case we don't need this line because the both have the same value."name"  is no longer needed and can be removed.

packages/ui-library/src/components/textarea/index.stories.ts
angsherpa456: Thanks for the hint, adjusted now for all the events.

JpunktWpunkt (approved):

PR #899, opened 2024-02-08
fix(ui-library): refactoring radio group to use slot instead of array…

davidken91:
… of options

### DIFF
--- a/packages/ui-library/src/components/forms/radio/radio-group/index.ts
+++ b/packages/ui-library/src/components/forms/radio/radio-group/index.ts
@@ -4,8 +4,7 @@ import { customElement, property } from 'lit/decorators.js';
 import { styleCustom } from './index.css';
 import { formDark, formLight } from '../../../../foundation/semantic-tokens/form.css';
 import { radioDark, radioLight } from '../../../../foundation/component-tokens/radio.css';
-import { InputSizesType, RadioOption } from '../../../../globals/types';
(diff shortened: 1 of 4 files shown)


ChristianHoffmannS2 (approved):

ChristianHoffmannS2:
LGTM. i am a bit unsure about the deleted required attribute, but we will tackle that entire validation topic soon in a different scope, so approved

packages/ui-library/src/components/forms/radio/radio-group/index.ts
```diff
@@ -59,15 +51,13 @@ export class BlrRadioGroup extends LitElement {
...
-      disabled: this.disabled || false,
-      readonly: this.readonly || false,
       checked: this.checked || false,
       error: this.hasError || false,
     });
 
-    const calculateOptionId = (label: string) => {
-      return label.replace(/ /g, '_').toLowerCase();
-    };
+    // const calculateOptionId = (label: string) => {
```
angsherpa456, line 58: Is this still required?

davidken91: No, I can remove it. I'm just not too sure what to do with this PR as a whole because I believe that @seemann was also working on a similar topic?

seemann: If you dont need this changes, then I would say that we should not merge it. If we do so the radio group will not really work anymore as I descripted in Slack.

PR #907, opened 2024-02-12
fix(ui-library): textarea label space alignment and counter alignment

RubirajAccenture:


packages/ui-library/src/components/forms/textarea/index.css.ts
```diff
@@ -73,7 +76,6 @@ export const { tokenizedLight: textAreaLight, tokenizedDark: textAreaDark } = re
       max-width: 100%;
       word-break: break-all;
       width: 100%;
```
davidken91, line 78: Do we need to add the background colour of white here like the ticket says? Something like this maybe? `        background-color: ${SurfaceFill.Default.Hover};`

davidken91 (commented):
On line 53 in the textarea.stories.ts file, you can replace `errorIcon: 'blr360` with `errorIcon: ''`. This will resolve another one of the acceptance criteria of the ticket. Then you can remove the unused `// @ts-expect-error todo` on line 456

davidken91:
Also, I can't seem to recreate this issue on the develop branch: _Currently in code the spacing between input and label is there, even if the hasLabel=false. In design this spacing is not there._ Maybe we need to align with Lars or Thorben on this

davidken91 (approved):

packages/ui-library/src/components/forms/textarea/index.ts
```diff
@@ -186,29 +189,24 @@ export class BlrTextarea extends LitElement {
             })
           : nothing}
       `;
-
+      //prettier-ignore
```
remmycat, line 192: I wanted to suggest adding an additional comment here to explain the issue, but I think I found a better solution, with which you can remove this again, see the next comment.

```suggestion

```

packages/ui-library/src/components/forms/textarea/index.ts
```diff
@@ -227,27 +225,20 @@ export class BlrTextarea extends LitElement {
             @blur=${this.handleBlur}
             @select=${this.handleSelect}
             @keyup=${this.updateCounter}
-          >
-${this.value}
-          </textarea
-          >
+          >${this.value}</textarea>
```
remmycat, line 228: I just found that the lit docs warn that you shouldn't put the text value as a child inside the textarea element in the first place: https://lit.dev/docs/templates/expressions/#invalid-locations

Instead it is suggested to pass it as `.value` attribute:

```suggestion
            .value=${this.value}
          ></textarea>
```

As a side effect, prettier shouldn't mess it up anymore too 🙂

yarn.lock
remmycat, line 1: As far as I can tell the yarn.lock shouldn't be changed, was this accidental?

When I check out the branch, reset the yarn.lock and run `yarn`, it doesn't create these changes.

I think you should remove the changes by resetting the yarn.lock to its state before your commit, unless you purposefully updated dependencies as part of this PR.

remmycat (approved):

davidken91 (approved):

PR #913, opened 2024-02-13
Feat/improve token structure v2

larserbach:
This PR includes a lot of changes on semantic tokens and system tokens (sys). The latter was previously referred to as "intermediary"

Also all css is updated to the new token names.


packages/ui-library/src/utils/range-slider-utils.ts
```diff
@@ -39,13 +39,13 @@ export const generateRangeBar = (
...
 
-    const activeDefaultColor = Forms.Slider.Track.Border.Default.Active.color;
-    const activeMuteColor = Forms.Slider.Track.Border.Default.Inactive.color;
+    const activeDefaultColor = Slider.Track.Border.Default.Active.color;
+    const activeMuteColor = Slider.Track.Border.Default.Inactive.color;
 
-    const disabledDefaultColor = Forms.Slider.Track.Border.Mute.Active.color;
-    const disbledMuteColor = Forms.Slider.Track.Border.Mute.Inactive.color;
+    const disabledDefaultColor = Slider.Track.Border.Mute.Active.color;
+    const disbledMuteColor = Slider.Track.Border.Mute.Inactive.color;
```
davidken91, line 48: Small typo here

packages/ui-library/src/foundation/semantic-tokens/form.css.ts
```diff
@@ -225,15 +228,15 @@ export const { tokenizedLight: formLight, tokenizedDark: formDark } = renderThem
...
       }
 
       &.md {
-        margin: ${MD.CaptionSlot.Margin};
+        margin: ${forms.captionslot.margin.md};
       }
 
       &.lg {
-        margin: ${MD.CaptionSlot.Margin};
+        margin: ${forms.captionslot.margin.md};
```
davidken91, line 239: I was about to point out that this could be a mistake, but I actually don't think we use `captionWrapper` any more

packages/ui-library/src/foundation/component-tokens/radio.css.ts
```diff
@@ -2,9 +2,9 @@ import { typeSafeNestedCss } from "../../utils/nested-typesafe-css-literals";
 
 import { renderThemedCssStrings } from "../_tokens-generated/index.pseudo.generated";
 
-export const { tokenizedLight: radioLight, tokenizedDark: radioDark } = renderThemedCssStrings((componentTokens, semanticTokens) => {
-  const { SM, MD, LG, Legend } = semanticTokens.Forms;
-  const { Radio, RadioGroup, FormLabel, CaptionComponent } = componentTokens.Forms;
+export const { tokenizedLight: radioLight, tokenizedDark: radioDark } = renderThemedCssStrings((componentTokens) => {
+  // const { } = semanticTokens.sem.forms;
```
davidken91, line 6: Can we remove this?

packages/ui-library/src/components/forms/toggle-switch/index.css.ts
```diff
@@ -14,9 +14,9 @@ export const styleCustom = typeSafeNestedCss/* css */ `
 
 export const { tokenizedLight: toggleSwitchLight, tokenizedDark: toggleSwitchDark } = renderThemedCssStrings(
   (componentTokens, semanticTokens) => {
-    const { ToggleSwitch, FormLabel } = componentTokens.Forms;
+    const { ToggleSwitch, FormLabel } = componentTokens.cmp;
```
davidken91, line 18: I haven't narrowed down why this is happening just yet, but I can see that the Toggle Switch icons colour is different to what's on develop. I spotted a few other things such as the icons being vertically aligned and the knob being the incorrect colour in several states, but I think they can be resolved in a separate ticket

packages/ui-library/src/components/forms/text-input/index.css.ts
```diff
@@ -174,62 +175,62 @@ export const { tokenizedLight: textInputLight, tokenizedDark: textInputDark } =
...
-      border: ${Input.Default.Rest.width} ${Input.Default.Rest.style} ${Input.Default.Rest.color};
-      border-radius: ${InputBorderRadius};
+      border: ${inputfield.container.border.default.rest.width} ${inputfield.container.border.default.rest.style} ${inputfield.container.border.default.rest.color};
+      border-radius: ${inputfield.container.borderradius};
       box-sizing: border-box;
 
       .blr-input-icon {
         position: relative;
-        color: ${Placeholder.Default.Rest};
+        color: ${inputfield.placeholder.textcolor.default.rest};
```
davidken91, line 184: A similar thing to what's happening on Toggle Switch is happening here. It looks like the icon colour is different to what's in Figma. Unless the design has changed

davidken91 (commented):
This probably goes beyond the scope of this ticket, but I just realised that we can remove `[this.size]: this.size,` from `iconClasses` in the following components: select, text-input and form-caption. This class is redundant since the component config size gets passed to the icon render function. It's by no means urgent to remove, but we might tidy it up later

packages/ui-library/src/utils/range-slider-utils.ts
larserbach: fixed

packages/ui-library/src/foundation/semantic-tokens/form.css.ts
larserbach: fixed it anyhow

packages/ui-library/src/foundation/component-tokens/radio.css.ts
larserbach: fixed

packages/ui-library/src/components/forms/toggle-switch/index.css.ts
larserbach: as discussed, figma was behind therefor the colors differed

packages/ui-library/src/components/forms/text-input/index.css.ts
larserbach: same as on toggle switch, the difference was expected and correct

Issue #774, opened 2024-01-12
NumberInput - fix accessibility violations


thrbnhrtmnn:
#### Description / User Story

The following errors and incomplete tests are thrown by the storybook accessibility addon and should be fixed:
- Violation: Buttons must have discernible text
- Violation: Form elements must have labels (placeholder="")
- Violation: Elements must meet minimum color contrast ratio thresholds (theme=dark)

----
#### Requirements
- [x] #534 

----
#### Acceptance Criteria
- [x] All errors and incomplete tests thrown by the accessibility addon are resolved:
   - [x] Violation: Buttons must have discernible text (readonly=false / unit=defined)
   - [x] Violation: Form elements must have labels (placeholder="")
   - [x] Violation: Elements must meet minimum color contrast ratio thresholds (theme=dark & unit=defined)

----
#### Background information
- ...

thrbnhrtmnn:
Approved

PR #917, opened 2024-02-14
fix 774 number input accessibility

remmycat:
Issue: #774 

The changes to the label were necessary to preserve the releation between the input and its label, both for the accessibility tree, and for standard browser behaviour (e.g. click on label focuses input)

### DIFF
--- a/packages/ui-library/src/components/form-label/index.stories.ts
+++ b/packages/ui-library/src/components/form-label/index.stories.ts
@@ -4,6 +4,8 @@ import { FormSizes, LabelVariants } from '../../globals/constants';
 import { BlrFormLabelType } from './index';
 import { BlrFormLabelRenderFunction } from './renderFunction';
 import { html } from 'lit-html';
+import { LitElement } from 'lit';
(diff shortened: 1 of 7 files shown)


packages/ui-library/src/components/internal-components/form-label/index.test.ts
```diff
@@ -14,16 +16,39 @@ const sampleParams: BlrFormLabelType = {
   variant: 'label',
 };
 
+// The label is not creating a shadow root itself, but errors if it is outside
+// of one. Thus, we're creating a helper component for the stories, that wraps it.
+class WrappedBlrLabel extends LitElement {
+  labelProps: BlrFormLabelType;
+
+  protected render() {
+    return html`${BlrFormLabelRenderFunction(this.labelProps)}`;
```
ChristianHoffmannS2, line 25: minor thing: these render functions already return html``-things. So this wrapping in html-tag-function is not needed

ChristianHoffmannS2 (approved):

ChristianHoffmannS2 (approved):

davidken91 (approved):

PR #926, opened 2024-02-16
Refactor event typing

remmycat:


ChristianHoffmannS2 (approved):

packages/ui-library/src/components/icon/index.ts
```diff
@@ -9,7 +9,15 @@ import { until } from 'lit-html/directives/until.js';
...
 import { TAG_NAME } from './renderFunction';
 import { ThemeType } from '../../foundation/_tokens-generated/index.themes';
+import { BlrClickEvent, createBlrClickEvent } from '../../globals/events';
 
+export type BlrIconEventHandlers = {
+  blrClick?: (event: BlrClickEvent) => void;
+};
+
+/**
+ * @fires blrClick Icon got clicked
```
angsherpa456, line 19: For consistency, just like above "Icon was clicked" might be better

packages/ui-library/src/components/icon/index.ts
```diff
@@ -20,13 +28,8 @@ export class BlrIcon extends LitElement {
...
 
-  // these are not triggered directly but allows us to map it internally and bve typesafe
-  @property() blrClick?: () => void;
-
-  protected handleClick = (event: Event) => {
-    this.dispatchEvent(
-      new CustomEvent('blrClick', { bubbles: true, composed: true, detail: { originalEvent: event } })
-    );
+  protected handleClick = (event: MouseEvent | KeyboardEvent) => {
+    this.dispatchEvent(createBlrClickEvent({ originalEvent: event }));
```
angsherpa456, line 32: Curly bracket can be removed

packages/ui-library/src/components/text-input/index.ts
```diff
@@ -14,7 +14,30 @@ import { BlrFormCaptionRenderFunction } from '../form-caption/renderFunction';
...
+  blrBlur?: (event: BlrBlurEvent) => void;
+  blrTextValueChange?: (event: BlrTextValueChangeEvent) => void;
+  blrSelect?: (event: BlrSelectEvent) => void;
+};
+
+/**
+ * @fires blrFocus TextInput received focus
+ * @fires blrBlur TextInput lost focus
+ * @fires blrTextValueChange TextInput value changed
+ * @fires blrSelect Text in input got selected
```
angsherpa456, line 39: Can you also change "blrSelect Text in input got selected" to "blrSelect Text in TextInput got selected" for consistency.

packages/ui-library/src/globals/events.ts
```diff
@@ -0,0 +1,79 @@
+export type BlrFocusEventDetail = {
+  originalEvent: FocusEvent;
+};
+export type BlrFocusEvent = CustomEvent<BlrFocusEventDetail>;
+export const BlrFocusEventName = 'blrFocus';
+export function createBlrFocusEvent(detail: BlrFocusEventDetail): BlrFocusEvent {
```
angsherpa456, line 6: Nice approach 👍

packages/ui-library/src/components/icon/index.ts
remmycat: It probably shouldn't be - `this.dispatchEvent()` returns a boolean, if I remove the brackets `handleClick` returns a boolean too, which doesn't make sense and could theoretically have an unexpected effect where handleClick is used.

angsherpa456: Aah, true, make sense, thanks for the insight.

angsherpa456 (dismissed):
LGTM

ChristianHoffmannS2 (approved):

angsherpa456 (approved):
LGTM

PR #929, opened 2024-02-16
fix(icons): auto generated icon type needs const array

remmycat:
This changes the `SizelessIcon` type from `string` to the union of valid icon names.


ChristianHoffmannS2 (approved):

angsherpa456 (approved):
LGTM

PR #937, opened 2024-02-20
feat(ui-components): implement events for number input

remmycat:
Notes:

- I tried writing tests for the other events too, but simulating the mouse and change events got rather complex - I couldn't get it working in the test environment. (I tried this: https://modern-web.dev/docs/test-runner/commands/#send-keys )
- I added the type hints for the properties because I noticed the number conversions weren't happening without them, which broke my code. The other code just always converted everything that might be a string into a number / boolean. I removed those explicit conversions, since this would sweep invalid types under the rug otherwise. With the type hint lit will convert the values for us.

Ticket: #481


packages/ui-library/src/components/number-input/index.stories.ts
JpunktWpunkt, line 1: image

This lines can be removed, because they are still there from line 266 and following

packages/ui-library/src/components/number-input/index.stories.ts
```diff
@@ -285,37 +285,47 @@ export default {
...
-    onChange: {
-      name: 'onChange',
+    //Events
+    blrNumberValueChange: {
+      name: 'blrNumberValueChange',
       description: 'Fires when the value changes.',
-      action: 'onChange',
+      action: 'blrNumberValueChange',
       table: {
         disable: false,
```
JpunktWpunkt, line 294: can also be removed

JpunktWpunkt, line 294: this can also be removed, because now we want to see the description

packages/ui-library/src/components/number-input/index.stories.ts
```diff
@@ -285,37 +285,47 @@ export default {
...
     },
-    onSelect: {
-      name: 'onSelect',
+    blrSelect: {
+      name: 'blrSelect',
       description: 'Fires when some text is selected.',
-      action: 'onSelect',
+      action: 'blrSelect',
       table: {
         disable: false,
```
JpunktWpunkt, line 303: l. 303 this can also be removed, because now we want to see the description

packages/ui-library/src/components/number-input/index.stories.ts
```diff
@@ -285,37 +285,47 @@ export default {
...
     },
-    onFocus: {
-      name: 'onFocus',
+    blrFocus: {
+      name: 'blrFocus',
       description: 'Fires when the component is focused.',
-      action: 'onFocus',
+      action: 'blrFocus',
       table: {
         disable: false,
```
JpunktWpunkt, line 312: l. 312 this can also be removed, because now we want to see the description

packages/ui-library/src/components/number-input/index.stories.ts
```diff
@@ -285,37 +285,47 @@ export default {
...
     },
-    onBlur: {
-      name: 'onBlur',
+    blrBlur: {
+      name: 'blrBlur',
       description: 'Fires when the component lost focus.',
-      action: 'onBlur',
+      action: 'blrBlur',
+      table: {
+        disable: false,
```
JpunktWpunkt, line 321: l. 321 this can also be removed, because now we want to see the description

faselbaum (dismissed):

faselbaum (approved):

davidken91 (approved):

PR #939, opened 2024-02-21
fix(781):Ensure Consistency between Storybook Documentation and Codebase

RubirajAccenture:
…ase Property Names

### DIFF
--- a/packages/ui-library/src/components/form-label/index.test.ts
+++ b/packages/ui-library/src/components/form-label/index.test.ts
@@ -9,10 +9,10 @@ import { LitElement } from 'lit';
 
 const sampleParams: BlrFormLabelType = {
   theme: 'Light',
-  labelText: 'Label text',
+  label: 'Label text',
   labelAppendix: 'Appendix txt',
   forValue: 'for_txt',
-  variant: 'label',
+  hasError: false,
 };
 
 // The label is not creating a shadow root itself, but errors if it is outside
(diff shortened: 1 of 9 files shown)


packages/ui-library/src/components/form-label/index.test.ts
```diff
@@ -102,12 +102,12 @@ describe('blr-form-label', () => {
...
       WrappedBlrFormLabelRenderFunction({
         ...sampleParams,
-        variant: 'error',
+        hasError: true,
       })
     );
     const blrLabel = querySelectorDeep('label.blr-form-label', element.getRootNode() as HTMLElement);
     const errorLabel = blrLabel?.getAttribute('class');
-    expect(errorLabel).to.contain('error');
+    expect(errorLabel).to.contain(true);
```
angsherpa456, line 110: May be it might be more reasonable to check for string [..to.contain('true')] rather than boolean OR if checking for boolean something like isTrue(...) might fit.

RubirajAccenture: Agreed. Changes done

angsherpa456 (approved):

JpunktWpunkt (changes requested):
image
image
image

The labelSize is now sizeVariant and line 23 can then be removed, the same then in lines 38/39 labelText ist now label , line 39 can be removed now, and also in line 60/61. variant is hasError and line 61 can be deleted too.

JpunktWpunkt (approved):
LGTM 🚀

angsherpa456 (approved):

JpunktWpunkt (commented):
LGTM 🚀

Issue #834, opened 2024-01-25
Number Input - Add Name Property


angelicahoyss:
#### Description / User Story

Property 'name' is currently missing in the Number Input component, even though it is displayed in Storybook.

----
#### Requirements
- [ ] #1019 

----
#### Acceptance Criteria
- [x] The component should have a 'name' property added.

thrbnhrtmnn:
Ticket will already be completed as part of #864 .

Issue #831, opened 2024-01-24
Number Input - Ensure Consistency between Storybook Documentation and Codebase Property Names


angelicahoyss:
#### Prerequisite

Name property is added to the component https://github.com/deven-org/B01LER-Kitchen/issues/834

#### Description / User Story

**As a** developer on the team,
**I want** to ensure consistency between the property names defined in the Storybook documentation for Number Input component
**So that** the Number Input component codebase accurately reflects the documented property names.

This user story outlines the need to maintain consistency between the documentation and the codebase regarding component property names. Each component is checked to ensure the property names in the codebase match the corresponding names in the Storybook documentation. **If the discrepancies are found, they are listed in[ this table](https://ts.accenture.com/:x:/r/sites/b01ler-core/_layouts/15/Doc.aspx?sourcedoc=%7Bb2cc9fb8-5c39-4049-9652-a4e6bbe09ef1%7D&action=editnew) on the page "Number Input".** Column Current Code Property Name describes how is the property currently named in the code. Column Desired Code Property Name is how the property should be named. Properties marked green don't need adjustment.

----
#### Acceptance Criteria

- [ ]  All Number Input component's property names that differ from Storybook documentation property names (listed in the above mentioned table) should be updated
- [ ] The updated property names should not introduce any unintended side effects or issues in the functionality of the components
- [ ] prependUnit should to be renamed in Storybook file (Documentation) to hasUnit, in addition to renaming the code property (this also includes changing this from a boolean to a radio with 3 options, changing the default to "none" and only displaying the unit prop in the prop panel if hasUnit is not "none")
- [ ] The properties in the props panel in Storybook are following the same order as in the Excel (i.e. the Accessibility section should come before the Technical Attributes section)
- [ ] The library builds successfully, and there are no reported issues related to the property name updates 
- [ ] [Props Excel](https://ts.accenture.com/:x:/r/sites/b01ler-core/Shared%20Documents/Documentation_Props.xlsx?d=wb2cc9fb85c3940499652a4e6bbe09ef1&csf=1&web=1&e=bhRYao) has been updated / comments have been resolved and props changed from red to black font color > @thrbnhrtmnn or @angelicahoyss can support here

JpunktWpunkt:
it would be nice to have this PR merged before we finished this [feat/481_number-input-events](https://github.com/deven-org/boiler/pull/937)

PR #940, opened 2024-02-21
fix(ui-library): ensure property naming number field (831)

JpunktWpunkt:
related to #831 and #834

### DIFF
--- a/packages/ui-library/src/components/input-field-number/index.stories.ts
+++ b/packages/ui-library/src/components/input-field-number/index.stories.ts
@@ -25,24 +25,26 @@ const defaultParams: BlrInputFieldNumberType = {
   value: undefined,
   decimals: 0,
   leadingZeros: 0,
-  prependUnit: true,
+  hasUnit: false,
   unit: 'kg',
+  unitPosition: 'prefix',
   step: 1,
   hasLabel: true,
   label: 'Label-text',
   labelAppendix: '(Appendix)',
   hasHint: false,
   hintMessage: 'This is a small hint',
-  hintIcon: 'blrInfo',
+  hintMessageIcon: 'blrInfo',
   disabled: false,
   readonly: false,
   required: false,
   hasError: false,
   errorMessage: '',
-  errorIcon: undefined,
-  inputFieldNumberId: 'test-id',
+  errorMessageIcon: undefined,
   stepIncreaseAriaLabel: '+',
   stepDecreaseAriaLabel: '\u2212',
+  inputFieldNumberId: 'inputFieldNumberId',
+  name: 'inputFieldNumber',
 };
 
 export default {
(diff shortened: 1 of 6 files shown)


RubirajAccenture (approved):

angsherpa456 (dismissed):

angsherpa456 (dismissed):

angsherpa456 (dismissed):

angsherpa456 (approved):

davidken91 (approved):

thrbnhrtmnn (changes requested):
Hey @JpunktWpunkt , the changes already look really good. I would just change a few minor things, before I will approve:
1. unitPosition should per default be "prefix" and we should not offer the "none" option
2. hasUnit please update the Description in the props panel to "Choose if component has a unit.". We should be a bit more consistent with similar props like hasLabel
3. Please align the order of the props in the props panel to the Excel (I think only unitPosition and the section "Accessibility" and "Technical Attributes" need to be adjusted)

thrbnhrtmnn (dismissed):
Storybook looks perfect now!

By the way, did you also solve https://github.com/deven-org/boiler/issues/834 with the implementation? If so, please let me know so I can close the other issue :-)

JpunktWpunkt:
> Storybook looks perfect now!
> 
> By the way, did you also solve #834 with the implementation? If so, please let me know so I can close the other issue :-)

Yes, I did :) #834 with the last commit when I check the order I saw that name was missing :)

packages/ui-library/src/components/input-field-number/index.ts
```diff
@@ -335,7 +337,9 @@ export class BlrInputFieldNumber extends LitElement {
                 @select=${this.handleSelect}
                 placeholder=${this.placeholder || ''}
               />
-              ${hasUnit ? html` <div class="${unitClasses}">${this.unit}</div> ` : nothing}
+              ${this.unitPosition === 'prefix' || this.unitPosition === 'suffix'
+                ? html` <div class="${unitClasses}">${this.unit}</div> `
```
angsherpa456, line 341: Please remove the extra space on both side of html

packages/ui-library/src/components/input-field-number/index.ts
```diff
@@ -257,40 +259,40 @@ export class BlrInputFieldNumber extends LitElement {
         [`${this.sizeVariant}`]: this.sizeVariant,
         [this.stepperVariant || 'split']: this.stepperVariant || 'split',
         'error-input': this.hasError || false,
-        'prepend': hasUnit && !!this.prependUnit,
+        // 'prepend': this.unitPosition === 'prefix' || this.unitPosition === 'suffix',
```
angsherpa456, line 262: Can this be removed?

packages/ui-library/src/components/input-field-number/index.stories.ts
```diff
@@ -132,7 +149,9 @@ export default {
       table: {
         category: 'Content / Settings',
       },
+      if: { arg: 'hasUnit', eq: true },
     },
+
```
angsherpa456, line 154: Please remove this additional line.

angsherpa456 (approved):
LGTM

thrbnhrtmnn (dismissed):

faselbaum (approved):

angsherpa456 (approved):
LGTM

PR #949, opened 2024-02-22
fix(ui-library): props alignment form label

RubirajAccenture:


JpunktWpunkt (approved):

davidken91 (commented):
It looks like a unit test is failing for form label. I think you just need to double check the error class that's being applied

packages/ui-library/src/components/form-label/index.stories.ts
```diff
@@ -56,13 +56,14 @@ export default {
     },
 
     hasError: {
+      name: 'hasError',
```
angsherpa456, line 59: Is this name required? I saw in some PR name property were removed for number of fields. Can you double check it?

packages/ui-library/src/components/form-label/index.stories.ts
```diff
@@ -56,13 +56,14 @@ export default {
...
       description: 'Choose if component has an error.',
-      options: LabelVariants,
-      control: { type: 'select' },
+      defaultValue: false,
       table: {
-        category: 'Validation',
+        category: 'Validations',
       },
     },
+
```
angsherpa456, line 66: This line can be removed.

JpunktWpunkt (dismissed):

ChristianHoffmannS2 (approved):

JpunktWpunkt (approved):

thrbnhrtmnn (approved):

PR #962, opened 2024-02-27
fix(ui-library, storybook): #823 radio - align prop names with design

### LINKED ISSUE #823: Radio - Ensure Consistency between Storybook Documentation and Codebase Property Names
#### Prerequisite

https://github.com/deven-org/B01LER-Kitchen/issues/824 is done and closed
https://github.com/deven-org/B01LER-Kitchen/issues/521 is done and closed

#### Description / User Story

**As a** developer on the team,
**I want** to ensure consistency between the property names defined in the Storybook documentation for Radio component
**So that** the Radio component codebase accurately reflects the documented property names.

This user story outlines the need to maintain consistency between the documentation and the codebase regarding component property names. Each component is checked to ensure the property names in the codebase match the corresponding names in the Storybook documentation. **If the discrepancies are found, they are listed in[ this table](https://ts.accenture.com/:x:/r/sites/b01ler-core/_layouts/15/Doc.aspx?sourcedoc=%7Bb2cc9fb8-5c39-4049-9652-a4e6bbe09ef1%7D&action=editnew) on the page "Radio".** Column Current Code Property Name describes how is the property currently named in the code. Column Desired Code Property Name is how the property should be named. Properties marked green don't need adjustment.

----
#### Acceptance Criteria

- [x]  All Radio component's property names that differ from Storybook documentation property names (listed in the above mentioned table) should be updated
- [x] The updated property names should not introduce any unintended side effects or issues in the functionality of the components
- [x] "araiaLabel" should be renamed to ariaLabel in Storybook file (Documentation)
- [x] The library builds successfully, and there are no reported issues related to the property name updates 
- [ ] [Props Excel](https://ts.accenture.com/:x:/r/sites/b01ler-core/Shared%20Documents/Documentation_Props.xlsx?d=wb2cc9fb85c3940499652a4e6bbe09ef1&csf=1&web=1&e=bhRYao) has been updated / comments have been resolved and props changed from red to black font color > @thrbnhrtmnn or @angelicahoyss can support here

bar-tay:
closes #823 

old branch at: https://github.com/deven-org/boiler/tree/fix/823-radio-prop-alignment_old

### DIFF
--- a/packages/ui-library/src/components/radio/index.stories.ts
+++ b/packages/ui-library/src/components/radio/index.stories.ts
@@ -2,6 +2,7 @@ import { BlrRadioType } from './index';
 import { BlrRadioRenderFunction } from './renderFunction';
 import { html } from 'lit-html';
 import { action } from '@storybook/addon-actions';
+import '../../index';
 
 // this loads the all components instances and registers their html tags
 import '../../index';
@@ -27,7 +28,7 @@ export default {
         category: 'Appearance',
       },
     },
-    size: {
+    sizeVariant: {
       name: 'sizeVariant',
       description: ' Choose size of the component. ',
       options: InputSizes,
@@ -80,7 +81,7 @@ export default {
         category: 'Content / Settings',
       },
     },
-    hintIcon: {
+    hintMessageIcon: {
       name: 'hintMessageIcon',
       description: 'Select an icon which is displayed in front of the hint message.',
       if: { arg: 'hasHint', eq: true },
@@ -127,7 +128,7 @@ export default {
       },
       if: { arg: 'hasError', eq: true },
     },
-    errorIcon: {
+    errorMessageIcon: {
       name: 'errorMessageIcon',
       description: 'Select an icon which is displayed in front of the error message.',
       table: {
@@ -146,7 +147,7 @@ export default {
       },
       control: { type: 'text' },
     },
-    optionId: {
+    radioId: {
       name: 'radioId',
       description: 'Unique identifier for this component.',
       table: {
@@ -160,28 +161,25 @@ export default {
       },
       control: { type: 'text', label: 'Radio' },
     },
-    onChange: {
-      name: 'onChange',
+    blrChange: {
       description: 'Fires when the value changes.',
-      action: 'onChange',
+      action: 'blrChange',
       table: {
         disable: false,
         category: 'Events',
       },
     },
-    onFocus: {
-      name: 'onFocus',
+    blrFocus: {
       description: 'Fires when the component is focused.',
-      action: 'onFocus',
+      action: 'blrFocus',
       table: {
         disable: false,
         category: 'Events',
       },
     },
-    onBlur: {
-      name: 'onBlur',
+    blrBlur: {
       description: 'Fires when the component lost focus.',
-      action: 'onBlur',
+      action: 'blrBlur',
       table: {
         disable: false,
         category: 'Events',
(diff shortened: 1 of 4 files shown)


davidken91:
There are two unit tests failing. Resolving it might be as simple as updating the class names used in the tests

packages/ui-library/src/components/radio/index.stories.ts
```diff
@@ -390,7 +388,7 @@ export const FormCaptionGroup = () => {
         ...args,
         hasHint: true,
         label: 'Hint message',
-        hintIcon: 'blr360',
+        hintMessageIcon: 'blr360',
```
thrbnhrtmnn, line 391: Icon should be blrInfo (probably will also be changed within #974 )

thrbnhrtmnn (commented):

packages/ui-library/src/components/radio/index.stories.ts
```diff
@@ -160,28 +158,28 @@ export default {
       },
       control: { type: 'text', label: 'Radio' },
     },
-    onChange: {
-      name: 'onChange',
+    blrChange: {
+      name: 'blrChange',
```
angsherpa456, line 162: name can be removed.

packages/ui-library/src/components/radio/index.stories.ts
```diff
@@ -160,28 +158,28 @@ export default {
...
+      action: 'blrFocus',
       table: {
         disable: false,
         category: 'Events',
       },
     },
-    onBlur: {
-      name: 'onBlur',
+    blrBlur: {
+      name: 'blrBlur',
```
angsherpa456, line 180: name can be removed

packages/ui-library/src/components/radio/index.stories.ts
```diff
@@ -2,6 +2,10 @@ import { BlrRadioType } from './index';
 import { BlrRadioRenderFunction } from './renderFunction';
 import { html } from 'lit-html';
 import { action } from '@storybook/addon-actions';
+import '../../index';
```
davidken91, line 5: You have to remove these lines. The app won't start with these duplicates

docs/ARCHITECTURE.md
```diff
@@ -1,25 +1,33 @@
 # Architecture
-The BO1LER, a fully customisable design system (DS), is developed using the [Lit](https://lit.dev/docs/) template and Figma. The design system serves as a core unit for UI components, styles, and guidelines, ensuring consistency and efficiency across our applications.
+
```
angsherpa456, line 2: Please revert back to the previous state where there was no new line added after every title. If you are using VS-Code then the new line might have been added by the prettier code formatter extension. Please disable that and remove the newly added line and save the file. Later on after committing your changes, you can re-enable the extension.

packages/ui-library/src/components/radio/index.stories.ts
bar-tay: done

packages/ui-library/src/components/radio/index.stories.ts
bar-tay: done

packages/ui-library/src/components/radio/index.stories.ts
```diff
@@ -160,28 +158,28 @@ export default {
...
+      action: 'blrChange',
       table: {
         disable: false,
         category: 'Events',
       },
     },
-    onFocus: {
-      name: 'onFocus',
+    blrFocus: {
+      name: 'blrFocus',
```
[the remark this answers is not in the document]
bar-tay: done

packages/ui-library/src/components/radio/index.stories.ts
bar-tay: done

packages/ui-library/src/components/radio/index.stories.ts
bar-tay: done

docs/ARCHITECTURE.md
bar-tay: done

angsherpa456:
Looks good, please adjust your commit number to one after rebasing to new develop branch then it should be good to be approved.

faselbaum (changes requested):
the js example app needs to be changed to reflect the property name changes.

davidken91:
> the js example app needs to be changed to reflect the property name changes.

Ok, I can update this. I just noticed that we use aria label in storybook but the prop is never actually used on the component itself. Should I remove this? I'm guessing we don't need aria-checked or aria-role since we're using the native radio input element

faselbaum:
@davidken91 

> Ok, I can update this. I just noticed that we use aria label in storybook but the prop is never actually used on the component itself. Should I remove this? I'm guessing we don't need aria-checked or aria-role since we're using the native radio input element

i would just leave them as is for the time being. but we should create a new issue so that we either in- or exclude them for all component stories.

faselbaum (approved):

PR #969, opened 2024-02-29
fix(ui-library): icon size behaviour (798)

ChristianHoffmannS2:


davidken91 (dismissed):

JpunktWpunkt (commented):
It looks, that the number input is a little bit broken now, same with the select, tab, text-input Component , also the icon page in storybook looks that they size has grown :)

ChristianHoffmannS2:
> It looks, that the number input is a little bit broken now, same with the select, tab, text-input Component , also the icon page in storybook looks that the size has grown :)

super, thanks!

well, the last one with the icons is fine for now, but the other ones should be addressed.

davidken91 (approved):

JpunktWpunkt (approved):

PR #974, opened 2024-03-01
fix(ui-library): align how hint & error messages are displayed in our component examples

RubirajAccenture:


thrbnhrtmnn (commented):
Hey @RubirajAccenture ,

3 ACs of the issue are not yet met when I look at the branch:
- [x]  TextInput: default error message icon is "undefined" in the Text Input
- [ ]  TextArea: default error message icon is "undefined" in the Text Area
- [ ]  Select: default error message icon is "undefined" in the Select

Also, if you added the tests to make sure, that the error message icon is always undefined per default, we should get errors for this. It seems to me that something is not working here.

RubirajAccenture:
> Hey @RubirajAccenture ,
> 
> 3 ACs of the issue are not yet met when I look at the branch:
> 
> * [ ]   TextInput: default error message icon is "undefined" in the Text Input
> * [ ]   TextArea: default error message icon is "undefined" in the Text Area
> * [ ]   Select: default error message icon is "undefined" in the Select
> 
> Also, if you added the tests to make sure, that the error message icon is always undefined per default, we should get errors for this. It seems to me that something is not working here.

Changes done

packages/ui-library/src/components/checkbox/index.test.ts
```diff
@@ -34,6 +34,20 @@ describe('blr-checkbox', () => {
     expect(className).to.contain('input-control');
   });
 
+  it(' error Icon is set to undefined', async () => {
```
angsherpa456, line 37: Please remove the extra space before the _error Icon_. Also would be nice to have it more contextual while reading by rephrasing it to something like **it('sets the error Icon to undefined')**.

packages/ui-library/src/components/checkbox/index.test.ts
```diff
@@ -34,6 +34,20 @@ describe('blr-checkbox', () => {
...
+      BlrCheckboxRenderFunction({
+        ...sampleParams,
+        errorIcon: undefined,
+      })
+    );
+
+    const checkBox = querySelectorDeep('textarea', element.getRootNode() as HTMLElement);
+    const errorIcon = checkBox?.getAttribute('errorIcon');
+
+    expect(errorIcon).to.be.equal(undefined);
```
angsherpa456, line 48: Could you may be check for undefined directly i.e. expect(errorIcon).beUndefined

packages/ui-library/src/components/radio-group/index.test.ts
```diff
@@ -38,6 +36,20 @@ describe('blr-radio-group', () => {
     expect(className).to.contain('input-control');
   });
 
+  it(' error Icon is set to undefined', async () => {
```
angsherpa456, line 39: Same as first comment.

packages/ui-library/src/components/radio/index.test.ts
```diff
@@ -54,6 +55,20 @@ describe('blr-radio', () => {
     expect(errorClassName).to.contain('error');
   });
 
+  it(' error Icon is set to undefined', async () => {
```
angsherpa456, line 58: same as first comment

packages/ui-library/src/components/select/index.test.ts
```diff
@@ -66,6 +66,20 @@ describe('blr-select', () => {
     expect(errorClassName).to.contain('error');
   });
 
+  it(' error Icon is set to undefined', async () => {
```
angsherpa456, line 69: same as first comment

packages/ui-library/src/components/text-input/index.test.ts
```diff
@@ -121,6 +121,20 @@ describe('blr-text-input', () => {
     expect(errorClassName).to.contain('error');
   });
 
+  it(' error Icon is set to undefined', async () => {
```
angsherpa456, line 124: same as first comment

packages/ui-library/src/components/textarea/index.test.ts
```diff
@@ -117,6 +117,20 @@ describe('blr-textarea', () => {
     expect(disabled).to.be.equal('');
   });
 
+  it(' error Icon is set to undefined', async () => {
```
angsherpa456, line 120: same as first comment

thrbnhrtmnn (commented):
Hey @RubirajAccenture ,

I forgot to list the Number Input in the issue. Can you also add the test to the Number Input?

Also I checked the Storybook pages and found an issue in Text Input and Select, which was not appearing in develop and seems to be introduced within your branch. When hasError=true, the whole input field gets bigger and also the icon disappeared. Could you check that?

thrbnhrtmnn:
As just discussed, here are the findings that were added to the ticket:

Open points:
- Radio: Form Caption Group Example has wrong icon for hint
- Radio Group: Form Caption Group Example has missing icon for error message
- Input Field Text: ErrorMessageIcon should be undefined per default
- Text Area: ErrorMessageIcon should be undefined per default
- Input Field Number: Form Caption Group Example needs to be fixed, the Has Hint example also has a hint message and the hint & error example has a hint incl. message and icon

packages/ui-library/src/components/radio-group/index.test.ts
thrbnhrtmnn: This still needs to be fixed

packages/ui-library/src/components/radio/index.test.ts
thrbnhrtmnn: Please fix this

packages/ui-library/src/components/select/index.test.ts
thrbnhrtmnn: This also still needs a fix

packages/ui-library/src/components/input-field-text/index.test.ts
```diff
@@ -121,6 +121,20 @@ describe('blr-input-field-text', () => {
     expect(errorClassName).to.contain('error');
   });
 
+  it(' error Icon is set to undefined', async () => {
```
thrbnhrtmnn, line 124: Please remove the space before "error Icon"

packages/ui-library/src/components/textarea/index.test.ts
thrbnhrtmnn: Please fix this one

thrbnhrtmnn (changes requested):
Hey @RubirajAccenture , thanks for the updates, here are my findings:
1. It seems to me that there is not yet a test for the input field number to make sure the error message icon is always undefined.
2. The Input Field Text and the Select currently get bigger in height when hasError=true, this should not happen.
3. In the Checkbox Docs, under Form Caption Group the hintMessageIcon should be blrInfo in the secon example (hint and error)

thrbnhrtmnn (approved):
Hey @RubirajAccenture , all points where fixed. I approve the PR, but I also found a very minor topic. If you have time you can fix it as well to improve consistency. I will quickly approve again in case you change something.

faselbaum (approved):

packages/ui-library/src/components/radio-group/index.test.ts
```diff
@@ -36,6 +36,20 @@ describe('blr-radio-group', () => {
...
 
+  it('error Icon is set to undefined', async () => {
+    const element = await fixture(
+      BlrRadioGroupRenderFunction({
+        ...sampleParams,
+        errorIcon: undefined,
+      })
+    );
+
+    const radioGroup = querySelectorDeep('textarea', element.getRootNode() as HTMLElement);
```
angsherpa456, line 47: Please target the correct element, not the textarea.

packages/ui-library/src/components/radio-group/index.test.ts
```diff
@@ -36,6 +36,20 @@ describe('blr-radio-group', () => {
...
+    const errorIcon = radioGroup?.getAttribute('errorIcon');
+
+    expect(errorIcon).to.be.undefined;
```
angsherpa456, line 50: Please adjust similar to prior tests.

packages/ui-library/src/components/radio/index.test.ts
```diff
@@ -54,6 +55,25 @@ describe('blr-radio', () => {
...
+        ...sampleParams,
+        hasError: true,
+        errorIcon: undefined,
+      })
+    );
+
+    const labelWrapper = querySelectorDeep('.label-wrapper', element.getRootNode() as HTMLElement);
+    const captionWrapper = querySelectorDeep('.caption-wraper', labelWrapper?.getRootNode() as HTMLElement);
+    const formCaptions = querySelectorAllDeep('.blr-form-caption', captionWrapper?.getRootNode() as HTMLElement);
+    const formCaptionError = querySelectorDeep('.blr-form-caption', formCaptions[1] as HTMLElement);
```
angsherpa456, line 70: Isn't it possible to target the error inside the form-caption directly just like in other cases? The querySelectorAllDeep is useful if both the hint and error needs to be tested. 
Note: hasHint can be set to false (if not already set in sampleParams) to assure that only one icon exist inside the form-caption.

packages/ui-library/src/components/radio/index.test.ts
```diff
@@ -54,6 +55,25 @@ describe('blr-radio', () => {
...
+
+    const errorIcon = formCaptionError?.getAttribute('errorIcon');
```
angsherpa456, line 72: Please adjust this like other test because this will ends up the line 74 always being truthy.

packages/ui-library/src/components/select/index.test.ts
```diff
@@ -66,6 +66,25 @@ describe('blr-select', () => {
...
+  it('error Icon is set to undefined', async () => {
+    const element = await fixture(
+      BlrSelectRenderFunction({
+        ...sampleParams,
+        hasError: true,
+        errorMessageIcon: undefined,
+      })
+    );
+
+    const labelWrapper = querySelectorDeep('.label-wrapper', element.getRootNode() as HTMLElement);
```
angsherpa456, line 78: Same as above packages/ui-library/src/components/radio/index.test.ts

packages/ui-library/src/components/textarea/index.test.ts
```diff
@@ -118,6 +118,20 @@ describe('blr-textarea', () => {
     expect(disabled).to.be.equal('');
   });
 
+  it('error Icon is set to undefined', async () => {
```
angsherpa456, line 121: Same as above packages/ui-library/src/components/radio/index.test.ts

faselbaum (approved):

angsherpa456:
Hi @RubirajAccenture, the current test code `
const errorIcon = formCaptionError?.getAttribute('errorIcon');
expect(errorIcon).to.not.exist;` 

will end up being a dead code because this test will always be truthy as there will never exist errorIcon attribute which getAttribute is trying to get. Please fix this as suggested in my previous comments.

Secondly, the title of the tests are still mismatch, please make it consistent with 
it('has error Icon set to undefined')..

PR #976, opened 2024-03-04
fix(ui-library, storybook): genericBlrComponentRenderer

### LINKED ISSUE #975: [Bug]: genericBlrComponentRenderer fails with empty props / attributes
### What is affected?

Other (please specify in the description)

## Affected:

- Package: @boiler/ui-library
- Component rendering in Storybook
- Component rendering in component tests
- Component rendering when using when using a different component's render function for composition purposes. E.g: https://github.com/deven-org/boiler/blob/c0de271eebb1204f9e32cf90d43c2c5dbb788ce1/packages/ui-library/src/components/slider-single-value/range-slider/index.ts#L51

## Description

There currently is an issue when passing an empty props or attributes argument to the `genericBlrComponentRenderer` function. Under these circumstances the resulting HTML template will be malformed and is unable to render properly.
This means that components without props will fail to render by default.

### Steps to reproduce

1. Pick a component without properties (or remove props from e.g. the divider component)
2. In the component story call the `<ComponentName>RenderFunction` with an empty object.
3. Run `yarn start` and navigate to the component in Storybook
4. Observe: Component rendering fails

**OR**

1. Pick a component without properties (e.g. remove props from divider)
2. In the component test call the `<ComponentName>RenderFunction` with an empty object.
3. Run `yarn test`
4. Observe: Test renderer fails

### Affected version(s)

@boiler/ui-library@0.0.1

OS: Sonoma 14.0 (23A344)
Browser: 122.0.6261.69 (Official Build) (arm64)
Node: v18.16.1
Package Manager: 4.1.0

### Code of Conduct

faselbaum:
- render util does not produce malformed HTML anymore when `props` or `htmlAttributes` are empty

Closes #975

### DIFF
--- a/packages/ui-library/src/utils/typesafe-generic-component-renderer.ts
+++ b/packages/ui-library/src/utils/typesafe-generic-component-renderer.ts
@@ -6,80 +6,96 @@ export const genericBlrComponentRenderer = <ComponentType extends { [s: string]:
   children?: TemplateResult<1>,
   htmlAttributes?: { [s: string]: string | boolean | number }
 ): TemplateResult<1> => {
-  const templateFragments: string[] = [];
+  let templateFragments: string[] = [];
   const values: unknown[] = [];
   const propEntries = Object.entries(props);
   const attrEntries = Object.entries(htmlAttributes || {});
 
-  // we will get rid of the dots later on by defining bindings within the component property decorator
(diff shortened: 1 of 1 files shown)


JpunktWpunkt (approved):
LGTM

PR #978, opened 2024-03-04
fix(ui-library): adding components prefix

davidken91:


packages/js-example-app/package.json
```diff
@@ -5,8 +5,8 @@
   "private": true,
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "build-index": "yarn run -T ejs src/index.ejs baseURL=. -o src/assets/index.html",
-    "build-index:prod": "yarn run -T ejs src/index.ejs baseURL=js-example-app -o src/assets/index.html",
+    "build-index": "yarn run -T ejs src/index.ejs componentsPrefix=adac baseURL=. -o src/assets/index.html",
```
angsherpa456, line 8: Is this intentional to have adac as prefix here?

davidken91: No actually, good spot. I'll replace this with blr

packages/ui-library/src/components/button-group/index.stories.ts
```diff
@@ -104,10 +104,7 @@ export const ButtonGroup = (
     })}
   `;
 
-  return html`<div class="wrapper">
-    ${BlrButtonGroupRenderFunction(params, contentButtons)}
-    <div class="wrapper"></div>
-  </div>`;
+  return html` <div class="wrapper">${BlrButtonGroupRenderFunction(params, contentButtons)}</div>`;
```
angsherpa456, line 107: Just a note: The extra space between html and div, can it be removed? Since the space in inside template literal, the extra space might get rendered as well but not sure.

packages/js-example-app/src/index.js
```diff
@@ -1,15 +1,16 @@
...
 const toggleLoadingButton = document.querySelector('#toggleLoadingState');
 const toggleDisabledState = document.querySelector('#toggleDisabledState');
 const logsContainer = document.querySelector('#logs');
 
-const blrButton = document.getElementsByTagName('blr-text-button')[0];
-const blrCheckbox = document.getElementsByTagName('blr-checkbox')[0];
-const blrSelect = document.getElementsByTagName('blr-select')[0];
-const blrTextInput = document.getElementsByTagName('blr-text-input')[0];
-const blrTextArea = document.getElementsByTagName('blr-textarea')[0];
+const button = document.getElementsByTagName(`${componentsPrefix}-text-button`)[0];
```
ChristianHoffmannS2, line 9: couldnt we use the already computed tag names here, from the render functions?
I.e. ButtonGroupTagName.

Maybe we should even include them in out ui-library/src/index.ts as exported variable?

PR #980, opened 2024-03-05
fix(ui-library): harmonised property names

JpunktWpunkt:
#876


davidken91 (dismissed):

thrbnhrtmnn (commented):

One small thing we could fix, while changing the index.stories.ts for the counter would be the text in the "Variant" section: In the text we are saying there is a "default" variant, but it is actually called "neutral".

angsherpa456 (dismissed):
LGTM

packages/ui-library/src/components/counter/index.ts
```diff
@@ -10,26 +10,26 @@ export class BlrCounter extends LitElement {
   static styles = [];
 
   @property() variant: CounterVariantType = 'neutral';
-  @property() current = 0;
-  @property() max = 0;
-  @property() size?: FormSizesType = 'md';
+  @property({ type: Number }) value = 0;
```
angsherpa456, line 13: Can it be more consistent because the type is defined for some and for some its missing.

packages/ui-library/src/components/counter/index.ts
```diff
@@ -10,26 +10,26 @@ export class BlrCounter extends LitElement {
...
   protected render() {
-    if (this.size) {
+    if (this.sizeVariant) {
       const dynamicStyles = this.theme === 'Light' ? [counterLight] : [counterDark];
 
       const classes = classMap({
         'blr-counter': true,
         [this.variant]: this.variant,
-        [`${this.size}`]: this.size,
+        [`${this.sizeVariant}`]: this.sizeVariant,
```
angsherpa456, line 25: For consistency, does this need to be in template literal, can't it be just like in line number 24?

thrbnhrtmnn (commented):
LGTM

packages/ui-library/src/components/counter/index.ts
JpunktWpunkt: I only do that in these case where lit analyze is barking. Apart from that, this is a topic we still need to talk about anyway. From consistent I agree with you.

thrbnhrtmnn (approved):

faselbaum (approved):

PR #981, opened 2024-03-05
fix(ui-library): harmonized property names

JpunktWpunkt:
#878

### DIFF
--- a/packages/ui-library/src/components/form-caption-group/index.ts
+++ b/packages/ui-library/src/components/form-caption-group/index.ts
@@ -10,12 +10,12 @@ import { TAG_NAME } from './renderFunction';
 export class BlrFormCaptionGroup extends LitElement {
   static styles = [formCaptionGroupStyle];
 
-  @property() size: FormSizesType = 'md';
+  @property() sizeVariant: FormSizesType = 'md';
 
   protected render() {
     const classes = classMap({
       'blr-form-caption-group': true,
-      [this.size]: this.size,
+      [this.sizeVariant]: this.sizeVariant,
     });
 
     return html`
--- a/packages/ui-library/src/components/checkbox/index.ts
+++ b/packages/ui-library/src/components/checkbox/index.ts
@@ -303,7 +303,7 @@ export class BlrCheckbox extends LitElement {
                 })}`
               : nothing}
             ${this.hasHint || this.hasError
-              ? BlrFormCaptionGroupRenderFunction({ size: this.size }, captionContent)
+              ? BlrFormCaptionGroupRenderFunction({ sizeVariant: this.size }, captionContent)
               : nothing}
           </div>
         </div>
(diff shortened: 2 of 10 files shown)


angsherpa456 (dismissed):
LGTM

thrbnhrtmnn (commented):
Also LGTM, we can approve and merge once the conflict is solved

packages/ui-library/src/components/slider-two-values/range-legend-min-max-slider/index.test.ts
```diff
@@ -3,50 +3,47 @@ import '@boiler/ui-library/dist';
...
-            expect(point).to.exist;
-        });
-        expect(allRangePoints.length).to.equal(sampleParams.list.length)
+    allRangePoints.forEach((point) => {
+      expect(point).to.exist;
     });
-})
+    expect(allRangePoints.length).to.equal(sampleParams.list.length);
+  });
+});
```
faselbaum, line 49: The formatting changes seem out of scope for this PR and could be removed.
What do you think? I'm curious, especially since i'm new to the team.
Should we just fix these things on the fly whenever we encounter them?

faselbaum (dismissed):

packages/ui-library/src/components/slider-two-values/range-legend-min-max-slider/index.test.ts
JpunktWpunkt: To be honest, I have no glue. I think this came when I solved the merge conflict  from develop. I don't touch this file in this way.

packages/ui-library/src/components/form-caption-group/index.ts
```diff
@@ -10,12 +10,12 @@ import { TAG_NAME } from './renderFunction';
...
   static styles = [formCaptionGroupStyle];
 
-  @property() size: FormSizesType = 'md';
+  @property() sizeVariant: FormSizesType = 'md';
 
   protected render() {
     const classes = classMap({
       'blr-form-caption-group': true,
-      [this.size]: this.size,
+      [this.sizeVariant]: this.sizeVariant,
```
davidken91, line 18: Do we know if the form caption component also needs 'size' to be renamed to 'sizeVariant'?

thrbnhrtmnn: Yes, we do, but it seems that there is not issue for this. @angelicahoyss , can you create one prop alignment issue for the Form Caption as well?

angelicahoyss: @thrbnhrtmnn @davidken91  Yes of course :) Here it is - https://github.com/deven-org/boiler/issues/1011

davidken91 (approved):

thrbnhrtmnn (approved):

Issue #970, opened 2024-03-01
Going Kebab-Case All The Way


thrbnhrtmnn:
### Description / User story

After implementing fixing the show code function #457 we realized that component names and attributes should not be written in camelCase, as in html they will just become lowercase without spacing. HTML is not case sensitive, but normally it is written in kebab-case.

With this issue we want to align our casing in code and switch to kebab-case.

We also will have a separate ticket to align the casing for the token system, which currently uses sentence case and camelCase plus occasionally "_" and "-". (#971 )

### Requirements / Prerequisites

- none

### Acceptance Criteria

- [ ] It has been clearly defined what parts of our code should use which casing (i.e. Component tags = kebab-case, Attributes = kebab-case etc.) > possibly everything can be changed to kebab-case
- [ ] We have updated the casing according to the new decision in the code
- [ ] We have updated the casing according to the new decision in the documentation

### Additional information

- ...

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #982, opened 2024-03-06
fix(ui-library): harmonised text area property naming (865)

JpunktWpunkt:
#865


packages/ui-library/src/components/textarea/index.stories.ts
```diff
@@ -36,12 +36,12 @@ const defaultParams: BlrTextareaType = {
...
   hintMessage: 'This is a small hint message',
-  hintIcon: 'blrInfo',
+  hintMessageIcon: 'blrInfo',
   arialabel: 'Text Area',
   name: 'Text Area',
-  showCounter: true,
+  hasCounter: true,
   theme: 'Light',
-  textareaId: '#textAreaId',
+  textAreaId: '#textAreaId',
```
davidken91, line 44: I this naming for TextArea. textAreaId seems ok to me, but does this mean that we have to change it everywhere in the app? For example, we may need to update the render function: https://github.com/deven-org/boiler/blob/develop/packages/ui-library/src/components/textarea/renderFunction.ts#L6, the label in our example JS app: https://github.com/deven-org/boiler/blob/develop/packages/js-example-app/src/index.ejs#L79 and the component type in index.ts: https://github.com/deven-org/boiler/blob/develop/packages/ui-library/src/components/textarea/index.ts#L256

packages/ui-library/src/components/textarea/index.stories.ts
```diff
@@ -395,7 +390,7 @@ TextArea.args = defaultParams;
 const argTypesToDisable = [
   'theme',
   'size',
```
thrbnhrtmnn, line 392: shouldn't this be sizeVariant?

packages/ui-library/src/components/textarea/index.stories.ts
thrbnhrtmnn: Hey @davidken91 , hey @JpunktWpunkt , the goal should be to be very consistent with our component names. 

In the documentation we are writing our components in Title Case, so it is Text Area. In code we decided to go with lowerCamelCase, so textArea would be correct. However we are looking into changing this with https://github.com/deven-org/boiler/issues/970 and to use kebab-case instead. Part of this issue is also to clearly define where to use which typecase.

Therefor I would say we should not update the examples you have linked and wait for #970.

thrbnhrtmnn (commented):

But I also found another smaller issues in Storybook, that we might could quickly change and fix: The order of the props in the props panel is not as defined in the Props Excel. For example the "states" props should be above "validation", then "accessibility" and then "technical attributes". Also "theme" should be the very first prop.

Also the props "warningLimitPer" and "warningLimitInt" should not be displayed when hasCounter is false. We may have discussed this before and found out that this is not possible, because also we should only show one of the props depending on the value of "warningLimitType".

JpunktWpunkt:
> LGTM in Storybook, cant evaluate the other parts though
> 
> But I also found another smaller issues in Storybook, that we might could quickly change and fix: The order of the props in the props panel is not as defined in the Props Excel. For example the "states" props should be above "validation", then "accessibility" and then "technical attributes". Also "theme" should be the very first prop.
> 
> Also the props "warningLimitPer" and "warningLimitInt" should not be displayed when hasCounter is false. We may have discussed this before and found out that this is not possible, because also we should only show one of the props depending on the value of "warningLimitType".

Yes we talked about that, storybook hasn't the opportunity to set more then one if condition in an easier way.

faselbaum (approved):

faselbaum (approved):

thrbnhrtmnn (approved):

PR #991, opened 2024-03-07
refactor(ui-library): change theming approach

### LINKED ISSUE #740: Theming - refactor current approach
#### Description / User Story

There are multiple topics, we can improve in our current theming approach:
- reducing build size / optimising code base by removing the <style> tag in each component
- allow more than 2 themes (ideally we have a class for the theme, and depending on the theme the components then get the right values

----
#### Requirements
- [ ] #713 is done

----
#### Acceptance Criteria
- [x] <style> tag has been removed from components
- [x] We now can have more than 2 themes in code
- [ ] New themes added by design are automatically added in code

----
#### Background information
- Currently for each theme a huge string will be created, which will also change dynamically the info in the style tags.
- A better solution would be to have one huge style string, but classes for light and dark.
- After this we can also check out CSS minifier to reduce the build size.
- Automating the adding of new theme can be done with the token studio meta data

ChristianHoffmannS2:
## What is included
- New css helpers to produce static styles from themed token sets (`ThemeIterator`)
- Removal of the old `renderThemedCssStrings` function
- A refactor of all existing css definitions to use the new iterator approach
- Removal of most dynamically inserted style tags in favor of lit's static `style` class field. See: [lit docs](https://lit.dev/docs/components/styles/#add-styles)

## What is not included
- Code generation to automatically include any new themes defined in tokens studio. Theme names and theme inclusion  are hard coded at the moment and need to be changed manually.

## Hints for reviewers
All changes are of similar nature. So it's not necessary to check each file meticulously. Reviewers should focus on finding visual deviations in Storybook when compared to the `develop` branch.

#### Systematic changes for each component:
- CSS definitions now always include all styles for all themes. This is facilitated through the use of the new `ComponentThemeIterator` / `SemanticThemeIterator` functions.
- Removal of most dynamically inserted style tags in favor of lit's static `style` class field. See: [lit docs](https://lit.dev/docs/components/styles/#add-styles). Exceptions are:
  - `form-label`s due to missing ShadowRoot
  - `range-slider` background gradient generation due to dynamic numerical calculations
- CSS selectors are repeated per iterator type (up to three times, once for: base styles without token values, semantic token iterator, component token iterator). This is to be expected.

## Issues found during review (Please add your findings here)
- [x] [`radio-group` layout is broken](https://github.com/deven-org/boiler/pull/991#issuecomment-2051664964)

Closes #740


thrbnhrtmnn (approved):

faselbaum:
### `radio-group` layout is broken

Before:
Screenshot 2024-04-12 at 14 24 00

After:
Screenshot 2024-04-12 at 14 24 13

packages/ui-library/src/components/form-caption-group/index.css.ts
```diff
@@ -1,28 +1,49 @@
...
+    const { forms } = sem;
 
-    &.lg {
-      gap: ${CaptionGroup.Container.ItemSpacing.LG};
-      margin: ${captionslot.margin.sm};
-    }
-  }
+    return typeSafeCss/*css*/ `
+      .blr-form-caption-group.${theme} {
+        &.sm {
```
angsherpa456, line 35: Since all the size has the same margin, it could be used as share styles:

`&.sm, &.md, &.lg {
margin: ${forms.captionslot.margin.sm};
}`

packages/ui-library/src/components/loader/index.ts
```diff
@@ -1,36 +1,38 @@
...
   @property() sizeVariant?: ActionSizesType = 'md';
   @property() variant?: FeedbackVariantType;
   @property() theme: ThemeType = 'Light';
 
   protected render() {
     if (this.sizeVariant) {
-      const dynamicStyles = this.theme === 'Light' ? [styleCustomLight] : [styleCustomDark];
+      const containerClasses = classMap({
+        'loader-container': true,
+        [`${this.sizeVariant}`]: this.sizeVariant,
```
angsherpa456, line 22: Please make this consistent with other i.e. without template literals. Same for line 23

packages/ui-library/src/components/divider/index.ts
```diff
@@ -1,30 +1,26 @@
...
 import { classMap } from 'lit/directives/class-map.js';
-import { dividerDark, dividerLight } from './index.css';
+import { staticStyles } from './index.css';
 import { TAG_NAME } from './renderFunction';
 import { ThemeType } from '../../foundation/_tokens-generated/index.themes';
 import { DividerVariationTypes } from '../../globals/types';
 import { LitElementCustom } from '../../utils/lit-element-custom';
 
 export class BlrDivider extends LitElementCustom {
+  static override styles = [staticStyles];
```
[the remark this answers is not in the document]
faselbaum: thanks for spotting this. forgot about removing it.
we should enable the `noImplicitOverride` flag in the tsc options and include the override keyword everywhere we do overrides.

new issue: #TBD

packages/ui-library/src/components/select/index.css.ts
```diff
@@ -1,63 +1,147 @@
...
-              padding: ${labelslot.padding.md};
-            }
-          }
-          &.lg {
-            & > .label-wrapper {
-              display: flex;
-              padding: ${labelslot.padding.lg};
-            }
-          }
+  slot {
```
angsherpa456, line 16: why is slot hidden? If there is no use for it, then we better remove the slot completely.

faselbaum: out of scope for this pr

thrbnhrtmnn (approved):
Looks good in Storybook

angsherpa456 (approved):
LGTM

Issue #965, opened 2024-02-29
Icon Link - remove component in code


thrbnhrtmnn:
### Description / User story

The icon link component is currently only existing in code and should be removed. Instances where it is used within other components should either use an icon, an input icon, an icon button or a stepper button instead.

### Requirements / Prerequisites

- [ ] #input icon

### Acceptance Criteria

- [ ] The icon link component has been removed from the code base
- [ ] Components that used the icon link now use another component instead

### Additional information

- ...

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #997, opened 2024-03-08
fix(ui-library): removed icon-link

JpunktWpunkt:
#965


faselbaum (commented):
I noticed that the IconLink is briefly mentioned here:

https://github.com/deven-org/boiler/blob/fix/965-remove-icon-link/README.md?plain=1#L251

faselbaum (approved):

PR #1013, opened 2024-03-12
fix(ui): camelCase to kebab-case (#970)

### LINKED ISSUE #970: Going Kebab-Case All The Way

angsherpa456:
## Objective
Ensure that all property names of our custom elements are reflected as kebab-case HTML attributes in the DOM.

## Solution: Simple inheritance & LitElement overrides
We created a base class `BoilerLitElement` that extends `LitElement` and overrides the static `createProperty()` method. This custom `createProperty()` method automatically converts the name of any class members marked with the `@property` decorator to a kebab-cased attribute in HTML. All our custom elements now extend `LitElementCustom` instead of `LitElement`.

This solution is simple, light-weight & works well with TypeScript's class inference features.

## Dropped approaches
We investigated other solutions to approach this issue but decided to drop them due to complexity or TypeScript incompatibility.

### Class mixin approach (dropped)
Class mixins as proposed in [Lit's documentation](https://lit.dev/docs/composition/mixins/) don't play well with TypeScript currently. They work partially and with workarounds but would introduce a lot of complexity to make them adaptable and future proof for our project. It might still be worth it to investigate the mixin pattern in the future but it is out of scope for this issue.

*References:*
- https://lit.dev/docs/composition/mixins/
- https://www.typescriptlang.org/docs/handbook/mixins.html
- https://www.typescriptlang.org/docs/handbook/mixins.html#alternative-pattern (Not investigated)
- https://github.com/microsoft/TypeScript/issues/30355

#### This is as far as we got during research

```typescript
import { LitElement, PropertyDeclaration } from 'lit';

// eslint-disable-next-line @typescript-eslint/no-explicit-any
export type Constructor<T> = new (...args: any[]) => T;

export type Constructable<TConstructorResult, TConstructorHost> = Omit<TConstructorHost, never> &
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  Constructor<TConstructorResult>;

// Have to declare an exported class interface to add additional properties to our mixin class.
// Results in non-ignorable error TS4094 otherwise.
export declare class AdditionalMixinProperties {
  static staticMember(): void;
  protected protectedMember(): void;
}

export const camelCaseToKebabCase = (str: string) => str.replace(/([a-z0-9])([A-Z])/g, '$1-$2').toLowerCase();

// The mixin factory function
export function kebabCaseAttributes<T extends Constructable<LitElement, typeof LitElement>>(constructor: T) {
  class KebabCased extends constructor {
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    constructor(...args: any[]) {
      super(...args);
    }

    static createProperty(name: PropertyKey, options?: PropertyDeclaration<unknown, unknown> | undefined): void {
      let customOptions = options;

      // generate the attribute name if it hasn't been defined or if it's disabled.
      if (typeof options?.attribute === 'undefined' || options?.attribute === true) {
        const attributeName = camelCaseToKebabCase(name.toString());
        customOptions = { ...options, attribute: attributeName };
      }

      return super.createProperty(name, customOptions);
    }
  }

  return KebabCased as Constructable<InstanceType<T> & AdditionalMixinProperties, T & typeof AdditionalMixinProperties>;
}

// Need to create an explicitly named intermediate class to extend from.
// Otherwise: 
// - static member inference won't work on the subclass.
// - Error TS4094 | Property ... of exported class expression may not be private or protected.
// This needs to be done for every class definition that wants to extend any number of mixins.
export class MixinBase extends kebabCaseAttributes(LitElement) {}

// Extend ^ instead of `kebabCaseAttributes(LitElement)` directly
export class CustomElement extends MixinBase {
  // access to iherited protected, private and static members works
}
```

This solution works on almost all fronts: proper type inference and autocompletion for statics and non-statics in mixin class implementation as well as classes extending the mixin.
It needs more polish to make it easier to adapt for other  future mixins. Eg. offer a utility function that creates mixin classes and does the proper type conversion through inference (if possible). It also has the caveat of having to create explicitly named intermediate classes as commented in the code above.

### Setting the attribute option in the `@property()` decorator
We discussed going the route of explicitly naming our attributes through the `@property({ attribute: 'some-attribute-name' })` syntax but decided against it since the current solution is succinct and easy to undo should we decide against it in the future.

closes #970


packages/ui-library/src/mixins/kebab-case-attributes.ts
```diff
@@ -0,0 +1,25 @@
+import type { PropertyDeclaration } from 'lit-element';
```
ChristianHoffmannS2, line 1: cool stuff :)

faselbaum (commented):
In addition to the other comments we could think about adding an import restriction rule to our `ui-library/.eslintrc.js` to prevent ourselves from accidentally using the standard LitElement.

```js
// .eslintrc.js
module.exports = {
  ...
  rules: {
    'no-restricted-imports': [
      'error',
      {
        patterns: [
          {
            group: ['lit'],
            importNamePattern: 'LitElement',
            message: `Don't use the default LitElement class. Import from /utils/boiler-lit-element instead`,
          },
        ],
      },
    ],
  },
  ...
}
```

packages/ui-library/src/utils/boiler-lit-element.ts
```diff
@@ -0,0 +1,31 @@
+import { LitElement, PropertyDeclaration } from 'lit';
+
+export const camelCaseToKebabCase = (str: string) => str.replace(/([a-z0-9])([A-Z])/g, '$1-$2').toLowerCase();
+
+export class BoilerLitElement extends LitElement {
```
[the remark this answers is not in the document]
angsherpa456: Thanks for the hint, adjusted now.

packages/ui-library/src/components/checkbox/index.ts
```diff
@@ -265,7 +266,8 @@ export class BlrCheckbox extends LitElement {
             ?checked=${this.currentCheckedState}
             ?indeterminate=${this.currentIndeterminateState}
             ?readonly=${this.readonly}
-            ?hasError=${this.hasError}
+            ?has-Error=${this.hasError}
+            ?has-Label=${this.hasLabel}
```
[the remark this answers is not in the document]
angsherpa456: Nice point, if required to be adjusted, we can do it in another ticket. But for now reverted to original state.

davidken91 (approved):

packages/ui-library/src/utils/lit-element-custom.ts
```diff
@@ -0,0 +1,31 @@
+import { LitElement, PropertyDeclaration } from 'lit';
+
+export const camelCaseToKebabCase = (str: string) => str.replace(/([a-z0-9])([A-Z])/g, '$1-$2').toLowerCase();
```
Jumace, line 3: this regular expression might cause problems with some specific attribute names, e.g. `XMLValue` would be transformed into `xmlvalue` but should be `xml-value`.

In this stackOverflow article is a different solution the covers more special cases:
https://stackoverflow.com/a/67243723

The code is not that well readable though, here a version with some better naming:
```js
const kebabize = (str) => str.replace(/[A-Z]+(?![a-z])|[A-Z]/g, (upperCaseGroup, isNotFirstCharacter) => (isNotFirstCharacter ? "-" : "") + upperCaseGroup.toLowerCase())
```

I would still add a comment above this function to explain what it is doing, something like: "This regEx goes through the whole input string and searches for a consecutive group of upper case letters, a group can also consist of only one element. For each group found the callback function is called, if the group starts with the first character of the input string only make the group lower case else add a "-" at the front of the group and then make everything lower case"

Jumace (commented):
This is a neat solution for the problem of the attributes.

Found three things over all.

- [ ] leftover `has-Error` instead of `hasError`
- [ ] the used regEx for creating kebab-case is a it flawed, posted an alternative solution
- [ ] Somewhere in the documentation it should be noted that only the `LitElementCustom` should be used

packages/ui-library/.eslintrc.js
```diff
@@ -11,4 +11,26 @@ module.exports = {
...
+    ],
+  },
+  overrides: [
+    {
+      files: ['src/utils/lit-element-custom.ts'],
+      rules: {
+        'no-restricted-imports': 'off',
+      },
+    },
+  ],
```
faselbaum, line 35: since for now this is the only exception it's probably simpler to just add a `// eslint-disable` comment in the referenced file instead. eslintrc overrides can easily get out of hand if not managed properly.

angsherpa456: Make sense, adjusted now.

packages/ui-library/src/utils/lit-element-custom.ts
angsherpa456: Thanks for the hint, now adjusted with the comment.

angsherpa456:
"_Somewhere in the documentation it should be noted that only the LitElementCustom should be used_"
Now the information is added in the architecture section.

faselbaum (commented):
What do we do about properties like these?

https://github.com/deven-org/boiler/blob/567930347fad84660cb98dd1b7bb7954125fb81e/packages/ui-library/src/components/input-field-text/index.ts#L47

They aren't camelCased even though they probably should be. But this might also produce unwanted behaviour with screen readers when kebab-cased. I presume we usually only use them to forward the values to elements further down the tree? We should probably investigate how screen readers behave when `aria` attributes are duplicated on the custom element and within the shadow dom of said element.

This might lead to screen readers reading the label or whatever twice.

Jumace (dismissed):

davidken91 (dismissed):

Jumace:
For the comment of @faselbaum I'd suggest to cover this in a new ticket. From my understanding should the current implementation not break anything, only if attributes like the `arialabel` would be written in camelCase it could break.
If you agree with my statement please get in touch with @thrbnhrtmnn to create a new ticket.

ChristianHoffmannS2:
> What do we do about properties like these?
> 
> https://github.com/deven-org/boiler/blob/567930347fad84660cb98dd1b7bb7954125fb81e/packages/ui-library/src/components/input-field-text/index.ts#L47
> 
> They aren't camelCased even though they probably should be. But this might also produce unwanted behaviour with screen readers when kebab-cased. I presume we usually only use them to forward the values to elements further down the tree? We should probably investigate how screen readers behave when `aria` attributes are duplicated on the custom element and within the shadow dom of said element.
> 
> This might lead to screen readers reading the label or whatever twice.

a few things here.
- property names should not match default html attribute names. (thats why it never was camelcased in the first place, it warns us against that)

- the input field is connected to a real label by id. i dont see a need for an additional arial label. and i am also not sure who accessibility tools would choose between those two label options.

i would ignore it for now, we will definitely stumble on this one during the accessibility epic.

angsherpa456:
> o be updated to reflect the changes as we

True, some of them are broken currently.

faselbaum (approved):

davidken91 (approved):

PR #1022, opened 2024-03-14
fix(ui-library, storybook): #875 align toggle prop names with design

bar-tay:
old branch at: https://github.com/deven-org/boiler/tree/fix/875-prop-alignment-toggle_old

### DIFF
--- a/packages/ui-library/src/components/toggle-switch/index.css.ts
+++ b/packages/ui-library/src/components/toggle-switch/index.css.ts
@@ -5,6 +5,9 @@ export const styleCustom = typeSafeNestedCss/* css */ `
   .blr-label-toggleswitch {
     display: flex;
     align-items: flex-start;
+    flex-direction: row;
+    justify-content: space-between;
+
 
     & input {
       all: initial;
@@ -32,14 +35,10 @@ export const { tokenizedLight: toggleSwitchLight, tokenizedDark: toggleSwitchDar
           }
         }
 
-        &.leading {
-          flex-direction: column;
-        }
 
-        &.trailing {
-          flex-direction: row;
-          justify-content: space-between;
-          align-items: center;
+        &.has-state-label {
+          flex-direction: column;
+        
         }
 
         & > .toggle-content-col {
@@ -85,6 +84,7 @@ export const { tokenizedLight: toggleSwitchLight, tokenizedDark: toggleSwitchDar
               position: absolute;
               top: 0;
               display: flex;
+              align-items:center;
             }
       
             & > .blr-form-label-inline {
@@ -139,7 +139,7 @@ export const { tokenizedLight: toggleSwitchLight, tokenizedDark: toggleSwitchDar
                 }
                 .toggle-icon > .toggle-icon-class {
                   color: ${ToggleSwitch.Control.AY11Icon.IconColor.Inactive.Pressed};
-                  color: pink;
+              
                 }
               }
 
--- a/packages/ui-library/src/components/toggle-switch/index.stories.ts
+++ b/packages/ui-library/src/components/toggle-switch/index.stories.ts
@@ -2,7 +2,7 @@
 import { BlrToggleSwitchType } from './index';
 import { BlrToggleSwitchRenderFunction } from './renderFunction';
 import { PureIconKeys } from '@boiler/icons';
-import { FormSizes, IconPositionVariant } from '../../globals/constants';
+import { FormSizes } from '../../globals/constants';
 import { Themes } from '../../foundation/_tokens-generated/index.themes';
 import { html } from 'lit-html';
 import '../../index';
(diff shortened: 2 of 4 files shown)


packages/ui-library/src/components/toggle-switch/index.css.ts
```diff
@@ -85,6 +85,7 @@ export const { tokenizedLight: toggleSwitchLight, tokenizedDark: toggleSwitchDar
               position: absolute;
               top: 0;
               display: flex;
+              align-items:center;
```
davidken91, line 88: Nice. I spotted this recently and it was bugging me 😅

packages/ui-library/src/components/toggle-switch/index.ts
```diff
@@ -91,32 +91,32 @@ export class BlrToggleSwitch extends LitElement {
     }
   };
 
-  @state() protected active = false;
+  @state() protected checkedd = false;
```
davidken91, line 94: Small typo here

davidken91 (commented):
Maybe we could add this component to our example js app to make sure that the events are functioning correctly. It's probably beyond the scope of this ticket though

davidken91 (commented):
I think this needs a little more work. Switching 'hasStateLabel' between trailing and leading seems to toggle the on / off labels visibility. We might need a better property name. Something like 'stateLabelAlignment' maybe

packages/ui-library/src/components/toggle-switch/index.ts
bar-tay: Will do

bar-tay: done

bar-tay:
@davidken91 I also thought it might make sense to add it to the js-example but probably in another ticket.

@thrbnhrtmnn what do you think about the current state. Should we rethink the behaviour and naming of the   "hasStateLabel"-prop?

thrbnhrtmnn:
Hey @bar-tay , the current state as desired yet. The hasStateLabel property should be a boolean, the default should be "FALSE", which would replace the current "trailing" value. The showStateLabel prop needs to be removed, as the state label will always be shown when hasStateLabel=TRUE (which replaces the "leading" value).

I will also add the ToggleSwitch to our ticket where we extend the JS Example Application. This should not be done as part of this issue :-)

packages/ui-library/src/components/toggle-switch/index.ts
```diff
@@ -125,7 +125,7 @@ export class BlrToggleSwitch extends LitElement {
         'disabled': this.disabled || false,
         'readonly': this.readonly || false,
         'hover': this.hovered || false,
-        'active': this.active || false,
+        'checkedd': this.checked || false,
```
angsherpa456, line 128: A small typo here.

bar-tay: done

angsherpa456:
Please fix the pipeline issue then it should be good to be approved.

bar-tay:
@thrbnhrtmnn could you check the current state.

davidken91:
I think it's almost ready for approval. I just had one more question. I'm not sure how it's supposed to look when hint is set to true. What we currently have might be correct, but I can't seem to find how it's supposed to look in Figma
Screenshot 2024-03-20 at 16 31 19

packages/ui-library/src/components/toggle-switch/index.ts
```diff
@@ -91,32 +91,32 @@ export class BlrToggleSwitch extends LitElement {
...
-      this.active = true;
+      this.checked = true;
       this.currentCheckedState = !this.currentCheckedState;
     }
   };
 
   protected handleRelease = () => {
     if (!this.disabled && !this.readonly) {
-      this.active = false;
+      this.checked = false;
```
thrbnhrtmnn, line 105: This has nothing to do with the checked property being renamed to active, right? Because it should be "active" now and not "checked".

packages/ui-library/src/components/toggle-switch/index.ts
```diff
@@ -125,7 +125,7 @@ export class BlrToggleSwitch extends LitElement {
         'disabled': this.disabled || false,
         'readonly': this.readonly || false,
         'hover': this.hovered || false,
-        'active': this.active || false,
+        'active': this.checked || false,
```
thrbnhrtmnn, line 128: same as above

thrbnhrtmnn (changes requested):
Not sure where this needs to be fixed, but currently the "hasLabel" prop also controls if the state labels are shown or not. This should not be the case. If hasStateLabel=true and hasLabel=false, the state labels should still be shown.

packages/ui-library/src/components/toggle-switch/index.stories.ts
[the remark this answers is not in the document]
bar-tay: @thrbnhrtmnn
1. I changed it so hinMessage is always visible but I wonder why hintMessage should be visible if hasHint is false  
2. Fixed

packages/ui-library/src/components/toggle-switch/index.ts
bar-tay: This is unrelated. The checked state is necessary here to display the the visual active state(black) of the component which was  missing before.

packages/ui-library/src/components/toggle-switch/index.stories.ts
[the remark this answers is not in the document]
bar-tay: Done

thrbnhrtmnn (dismissed):
Storybook looks perfect now

JpunktWpunkt (dismissed):

packages/ui-library/src/components/toggle-switch/index.ts
```diff
@@ -92,32 +92,32 @@ export class BlrToggleSwitch extends LitElementCustom {
     }
   };
 
-  @state() protected active = false;
+  @state() protected checked = false;
```
faselbaum, line 95: Perhaps we should rename this to `pressed` so that it doesn't get confused with the checked state. This would also match the token system nomenclature.

packages/ui-library/src/components/toggle-switch/index.ts
```diff
@@ -126,7 +126,7 @@ export class BlrToggleSwitch extends LitElementCustom {
         'disabled': this.disabled || false,
         'readonly': this.readonly || false,
         'hover': this.hovered || false,
-        'active': this.active || false,
+        'active': this.checked || false,
```
faselbaum, line 129: Perhaps we should rename the class to `pressed`. The meaning of `active` can be a bit ambiguous at times.

PR #1023, opened 2024-03-14
fix(ui-library): props alignment form caption

RubirajAccenture:


davidken91 (approved):
You need to add the new 'hasIcon' prop to storybook. I would recommend doing it on line 46: https://github.com/deven-org/boiler/pull/1023/files#diff-4fd2a76c547f390341fc08e229facc46bf9c01c7d2cad98a90c769ef90971b74R46
Something like this: 
`    hasIcon: {
      description: 'Choose if component has an icon.',
      defaultValue: true,
      control: {
        type: 'boolean',
      },
      table: {
        category: 'Content / Settings',
      },
    },`
    Then adding it again in `const args =` on line 107:
    https://github.com/deven-org/boiler/pull/1023/files#diff-4fd2a76c547f390341fc08e229facc46bf9c01c7d2cad98a90c769ef90971b74R107

davidken91 (commented):
I would recommend renaming the content heading in form-caption index.stories.ts to 'Content / Settings' with an 's' to keep it inline with the other storybook files

packages/ui-library/src/components/textarea/index.ts
```diff
@@ -186,7 +186,7 @@ export class BlrTextarea extends LitElement {
           ? BlrFormCaptionRenderFunction({
               variant: 'error',
               theme: this.theme,
-              size: this.sizeVariant,
+              sizeVariant: this.sizeVariant,
               message: this.errorMessage,
```
davidken91, line 190: We have a unit test failing for the textarea. This is because the unit test expects the textarea hint icon to be visible. We're currently not using the `hasIcon: true` property here so it will be undefined and therefore will not show. The way it currently works for formCaption is that `hasIcon` is an optional property. If it's not provided, it will be hidden. I was wondering if we should make it mandatory? If we do make it mandatory, would this mean that each component that uses formCaption should have the ability to toggle the icon on and off? I think it's something that we need to consider and discuss before we merge anything

packages/ui-library/src/components/form-caption/index.stories.ts
thrbnhrtmnn, line 1: In Storybook I like to see the following changes:
- remove size from props panel
- sizeVariant should be in the "Appearance section" (below "variant"), have a radio as a control, the description "Choose size of the component." (without "-" at the end) and the default should be "SM"
- icon property should work and the selected icon should be shown in the default story

packages/ui-library/src/components/form-caption/index.ts
```diff
@@ -16,39 +16,40 @@ export class BlrFormCaption extends LitElement {
...
+        this.sizeVariant.toUpperCase(),
       ]).toLowerCase() as SizesType;
 
       return html`
         <style>
           ${dynamicStyles}
         </style>
         <div class=${classes}>
-          ${this.icon
+          ${this.hasIcon
```
thrbnhrtmnn, line 52: I think this is wrong. The icon prop should still be there.

But what is missing that we want to newly introduce the "hasIcon" property in a similar way how we have it in the Input Field Text (where it is still named showInputIcon"). If this new property is false, there should not be any icon in the component example story and the icon prop should also be hidden in the prop panel.

Would be cool if you can add this.

packages/ui-library/src/components/textarea/index.ts
thrbnhrtmnn: The idea is, that every component that can display an optional Icon would in the future have an hasIcon prop to toggle if the component has an icon or not. If hasIcon is true, we also will have the icon prop to define which icon is shown. This logic should already implemented this way in the input field text, though the hasIcon prop is named showInputIcon instead (will be fixed with the prop alignments).

We discussed this when we went over all the props with Andelka, Lars and Christian. @ChristianHoffmannS2 maybe you have another opinion now or would this still be the preferred way?

thrbnhrtmnn (changes requested):

packages/ui-library/src/components/form-caption/index.stories.ts
RubirajAccenture: Done

packages/ui-library/src/components/form-caption/index.ts
RubirajAccenture: I have changed it to showIcon but it just means that we have changed the name from hasIcon to ShowIcon.

davidken91 (commented):
This seems to work well for me. The caption icon will appear by default unless the user sets the icon to undefined. Unless we wanted to add an additional prop to each consuming component called 'showCaptionIcon'. Any thoughts @thrbnhrtmnn / @ChristianHoffmannS2 ?

packages/ui-library/src/components/form-caption/index.ts
thrbnhrtmnn: Looking at the newest version in Storybook, this seems to work correct now. But the property name should still be "hasIcon" instead of "showIcon". Can you change that?

Another small issue I found is, that the "message" prop should also be within the "Content / Settings" section in the prop panel. Currently it is placed in its own section called "Content / Setting". It should be placed before the "hasIcon" prop.

thrbnhrtmnn:
@davidken91 I think I understand the issue now. If we add a "hasIcon" prop to the Form Caption, we would also have to add a  similar prop to all components that consume the Form Caption, right?

I am not sure if this is the best way to do this. My goal is just to align the behaviour how we show or hide icons for all components, if it makes sense. But maybe there are also reasons to not align this behaviour. 

Currently the Button Text and the Input Field Text component both have a prop to show/hide the icon and another prop to choose the icon. 

Other components like Icon, Button Icon and Select do not have such a prop. This makes sense as these components also always need to have an icon.

I would suggest that we either add a hasIcon prop also to the FormCaption and all components that consume the Form Caption OR that we remove the hasIcon prop from Button Text and Input Field Text as well. We should do what is less effort.

davidken91 (approved):

packages/ui-library/src/components/form-caption/index.stories.ts
```diff
@@ -44,12 +45,20 @@ export default {
...
+    },
+
     icon: {
       options: [undefined, ...PureIconKeys],
       control: { type: 'select' },
+      defaultValue: 'blr360',
       description: 'Select an icon which is displayed inside of the input.',
       table: {
-        category: 'Content / Setting',
+        category: 'Content / Settings',
```
thrbnhrtmnn, line 61: You fixed an issue I did not even see at first here. "Content / Settings" is the correct name for the category. Could you also fix it for the message prop, so message, and icon are in the same category. Please also check, that the order is correct. First message, then icon.

packages/ui-library/src/components/form-caption/index.stories.ts
```diff
@@ -44,12 +45,20 @@ export default {
       },
     },
 
+    showIcon: {
+      description: 'Choose if component has an icon.',
+      defaultValue: true,
+      control: { type: 'boolean' },
+      table: { category: 'Content / Settings' },
+    },
+
```
thrbnhrtmnn, line 54: As discussed, the showIcon prop should not have a function in the form caption anymore and therefor can be removed here

thrbnhrtmnn (changes requested):

packages/ui-library/src/components/form-caption/index.stories.ts
thrbnhrtmnn, line 1: Could you also change the icon in line 160 to blrInfo?

thrbnhrtmnn (approved):

davidken91 (approved):

PR #1031, opened 2024-03-20
fix(ui-library): remove margin in error state

### LINKED ISSUE #945: Input Field Text, Input Field Number, Select & Text Area - remove margin in error state if no caption is defined
### Description / User story

Currently there is a margin in the error state of the Input Field Text and Text Area, even if no caption is defined. This should not be the case.

Also the default for errorMessageIcon should be undefined for all 3 components.

- none

### Acceptance Criteria

- [ ] Text Area: There is no margin between the Input and the Caption, when no Caption is defined
- [ ] Checkbox: There is no margin between the checkbox and the Caption, when no Caption is defined
- [ ] Select: There is no margin between the Select and the Caption, when no Caption is defined
- [ ] Input Field Text: There is no margin between the Input and the Caption, when no Caption is defined
- [ ] Input Field Number: There is no margin between the Input and the Caption, when no Caption is defined
- [ ] The default of the errorIconMessage is undefined in Text Area, Input Field Text and Input Field Number
- [ ] A test has been implemented to make sure the errorIconMessage does not switch back to defined in the future for all of these 3 components

### Additional information

- The same issue also appears in other input components, these will be fixed within the following issues:
   - #589
   -  #602
   - #754

### Code of Conduct

angsherpa456:
closes #945


angsherpa456:
New PR created after new develop version was introduced.

PR #1036, opened 2024-03-21
fix(ui-library, figma-design-tokens): #1014 create & use tokens for inputfield icon

### LINKED ISSUE #1014: Select & Input Field Text - add new tokens for icon and align design and code
### Description / User story

After discussing #567 in todays design review we decided to move this issue to after release 1.0 and instead focus on aligning design and code for the icons in the select and the input field text components without introducing a new component.

The solution would therefor just use the existing icon component and apply color and size tokens which will be defined by design.

The whole topic will be revisited with #567 in the future (after release 1.0) to identify and research if we need a separate input icon component which might then probably replace the current solution.

This ticket also solves #487 so this ticket has been closed.

- none

### Acceptance Criteria

Select:
- [x] Design: The icon in the select uses its own color and sizing tokens.
- [x] Code: The icon in the select uses the defined color and sizing tokens from design.
- [x] Design: In error state the error icon has been removed. The chevron icon is always visible.
- [x] Code: In error state the error icon has been removed. The chevron icon is always visible.
- [x] Design & Code: The icon should be overwriteable

Input Field Text:
- [x] Design: The error icon inside the input field has been removed
- [x] Code: The error icon inside the input field has been removed
- [x] Design: In error state the icon part uses the newly defined color and sizing tokens
- [x] Code: In error state the icon part uses the defined color and sizing tokens from design
- [x] Design & Code: The icon should be overwriteable

### Additional information
- ...

### Code of Conduct

faselbaum:
Closes #1014


packages/figma-design-tokens/input/tokens/$themes.json
```diff
@@ -2344,4 +2348,4 @@
     },
     "group": "Sizes"
   }
-]
+]
```
angsherpa456, line 2351: Can this 'no newline at end of file' can be resolved. Please check in other files as well.

packages/ui-library/src/components/select/index.ts
```diff
@@ -50,7 +49,7 @@ export class BlrSelect extends LitElementCustom {
   @property() hasLabel?: boolean;
   @property() label!: string;
   @property() disabled?: boolean;
-  @property() sizeVariant?: FormSizesType = 'md';
+  @property() sizeVariant: FormSizesType = 'md';
```
angsherpa456, line 52: Note: right now in the code-base, sizeVariant property is mixed up between being optional and mandatory. We should make it a consistent.

packages/ui-library/src/components/select/index.ts
```diff
@@ -222,10 +208,10 @@ export class BlrSelect extends LitElementCustom {
                 })}
               </select>
             </div>
-            ${this.renderIcon(iconClasses)}
+            ${this.renderIcon()}
           </div>
           ${(this.hasHint && this.hintMessage) || (this.hasError && this.errorMessage)
-            ? BlrFormCaptionGroupRenderFunction({ sizeVariant: this.sizeVariant }, getCaptionContent())
+            ? BlrFormCaptionGroupRenderFunction({ sizeVariant: this.sizeVariant }, captionContent)
```
angsherpa456, line 214: In one of my recent previous PR I changed this to be a getCaptionContent function rather than just a const captionContent otherwise the captionContent const is always populated regardless of it gets used or not. Is there any reason why you changed it back?

faselbaum: Sorry for the oversight. Will change it to it to a non-inline function. I changed it because creating an inline function on every render costs time too. I converted it out of reflex without thinking about if it's actually more performant in this case 😅

packages/ui-library/src/components/select/index.ts
faselbaum: I agree 👍 

new issue: #TBD

packages/figma-design-tokens/input/tokens/$themes.json
faselbaum: this comes from token studio. we will always have this occur when tokens are changed on design side. our formatter isn't configured to run on token input files.

angsherpa456 (approved):
LGTM

PR #1039, opened 2024-03-22
fix(storybook): align props

bar-tay:


faselbaum (dismissed):

packages/ui-library/src/components/radio-group/index.stories.ts
```diff
@@ -230,18 +228,17 @@ export const BlrRadioGroup = (params: BlrRadioGroupType) => BlrRadioGroupRenderF
...
 } = {
   theme: 'Light',
-  size: 'md',
-  // direction: 'horizontal',
-  hasGroupLabel: true,
-  groupLabel: 'Group Label Text',
+  sizeVariant: 'md',
+  direction: 'horizontal',
+  hasLegend: true,
+  legend: '"Legend-text"',
```
angsherpa456, line 238: Is double quote inside the single quote intentional?

packages/ui-library/src/components/radio-group/index.ts
```diff
@@ -19,79 +19,80 @@ export class BlrRadioGroup extends LitElement {
...
-    if (!this.size) {
+    if (!this.sizeVariant) {
       return null;
     }
     const dynamicStyles = this.theme === 'Light' ? [formLight, radioLight] : [formDark, radioDark];
 
     const legendClasses = classMap({
       'blr-legend': true,
-      [`${this.size}`]: this.size,
+      [`${this.sizeVariant}`]: this.sizeVariant,
```
angsherpa456, line 49: Right now we have a mixed implementation of [`${this.sizeVariant}`] and [this.sizeVariant]. Since we already get what we want from the second part, we should convert [`${this.sizeVariant}`] to [this.sizeVariant].

packages/ui-library/src/components/radio-group/index.ts
```diff
@@ -19,79 +19,80 @@ export class BlrRadioGroup extends LitElement {
...
       'error': this.hasError || false,
     });
 
     const legendWrapperClasses = classMap({
       'blr-legend-wrapper': true,
-      [`${this.size}`]: this.size,
+      [`${this.sizeVariant}`]: this.sizeVariant,
```
angsherpa456, line 55: Right now we have a mixed implementation of [`${this.sizeVariant}`] and [this.sizeVariant]. Since we already get what we want from the second part, we should convert [`${this.sizeVariant}`] to [this.sizeVariant].

packages/ui-library/src/components/radio-group/index.ts
```diff
@@ -19,79 +19,80 @@ export class BlrRadioGroup extends LitElement {
...
     });
 
     const classes = classMap({
-      [`${this.size}`]: this.size,
+      [`${this.sizeVariant}`]: this.sizeVariant,
```
angsherpa456, line 59: Right now we have a mixed implementation of [`${this.sizeVariant}`] and [this.sizeVariant]. Since we already get what we want from the second part, we should convert [`${this.sizeVariant}`] to [this.sizeVariant].

packages/ui-library/src/components/radio-group/index.ts
```diff
@@ -19,79 +19,80 @@ export class BlrRadioGroup extends LitElement {
...
       disabled: this.disabled || false,
       readonly: this.readonly || false,
       checked: this.checked || false,
       error: this.hasError || false,
+      [`${this.direction}`]: this.direction,
```
angsherpa456, line 64: Please adjust [`${this.direction}`] to [this.direction].

packages/ui-library/src/components/radio-group/index.stories.ts
bar-tay: done

packages/ui-library/src/components/radio-group/index.stories.ts
```diff
@@ -277,13 +274,13 @@ const defaultParams: BlrRadioGroupType & {
...
   groupErrorMessage: '',
-  groupErrorIcon: undefined,
+  groupErrorMessageIcon: undefined,
   ariaLabel: 'Radio Group',
   radioGroupId: 'Radio Group',
   name: 'Radio Group ',
-  onChange: () => action('onFocus'),
-  onFocus: () => action('onFocus'),
-  onBlur: () => action('onBlur'),
+  blrChange: () => action('onFocus'),
```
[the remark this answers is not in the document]
bar-tay: done

packages/ui-library/src/components/radio-group/index.ts
bar-tay: done

packages/ui-library/src/components/radio-group/index.ts
bar-tay: done

packages/ui-library/src/components/radio-group/index.ts
bar-tay: done

packages/ui-library/src/components/radio-group/index.ts
bar-tay: done

angsherpa456:
Looks good, please first rebase to the new develop branch and squash commits to a single commit.

davidken91 (commented):
For the vertical variant of radio group, I think that the radio buttons and labels need to be inline like what we see in Figma
Screenshot 2024-03-28 at 12 18 27

davidken91:
A couple of small things I noticed.

1. By default, the label element comes with a small amount of padding to the left and right of it. By setting padding to 0 on the legend element (line 44 in radio.css.ts) then the legend will be flush with the radio inputs like in the design. This will work on both horizontal and vertical layouts
2. It looks like we're not using index.css.ts within the radio group directory. All of the styles associated with radio group come from either form.css or radio.css
3. It looks like we're not using the input control margin that's provided in figma. It only appears in md and lg. It is set to 0 on sm. Instead of using margin, we could simply add `align-items: center;` to .blr-radio on line 141 in radio.css.ts. That will automatically vertically align everything.
4. There are some notes in the radio.css.ts file saying ` _FIX_: "We need to differentiate between horizontal and vertical gaps, but this seems all wrong anyhow";` for now, the gaps look correct, but we need to double check this with design.
5. In the Figma file, I can see that we have individual hints per radio input. In storybook, we only have one global hint at the bottom. We need to double check what the functionality should be.

These issues are probably way outside of the scope of this ticket, but I think we may need a separate ticket to address these. Let me know if you have any thoughts @thrbnhrtmnn

faselbaum (approved):

davidken91 (approved):

PR #1060, opened 2024-03-27
fix(ui-library): remove margin in error state (#945)

### LINKED ISSUE #945: Input Field Text, Input Field Number, Select & Text Area - remove margin in error state if no caption is defined

angsherpa456:
closes #945


JpunktWpunkt (approved):

packages/ui-library/src/components/input-field-text/index.ts
```diff
@@ -238,8 +238,8 @@ export class BlrInputFieldText extends LitElementCustom {
                 )}`
               : nothing}
           </div>
-          ${this.hasHint || this.hasError
-            ? BlrFormCaptionGroupRenderFunction({ sizeVariant: this.sizeVariant }, captionContent)
+          ${(this.hasHint && this.hintMessage) || (this.hasError && this.errorMessage)
```
davidken91, line 241: I'm not suggesting this logic is wrong, but I just want to make sure that I understand it correctly. If hasHint is true and a hintIcon is set, but no hintMessage exists, are we hiding the hintIcon? That makes sense to me, but I just want to clarify

angsherpa456: Yes, that's correct, we are hiding the formCaption entirely because that is where the margin was coming. HintIcon/message & ErrorIcon/message is located inside the formCaption.

Issue #1024, opened 2024-03-14
Fix change events in components so they also include information on what was changed


thrbnhrtmnn:
### Description / User story

Currently we have change events in the following components:

Event | Components
-- | --
blrChange | SliderSingleValue, RadioGroup, ToggleSwitch, Radio, SliderTwoValues, TabBar
blrCheckedChange | Checkbox
blrNumberValueChange | NumberInput
blrSelectedValueChange | Select
blrTextValueChange | TextArea, TextInput

In the select we encountered the issue, that the event informs users that something changed, but it misses the information on what was changed. Probably similar issues exist in the other components as well.

With this task we want to fix this and also include the information on what was changed within the change events.

### Requirements / Prerequisites

- none

### Acceptance Criteria

- [ ] All components with change events also inform users/event listeners on what was changed when the event fires

### Additional information

- related: #972 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1061, opened 2024-03-27
fix(ui-library): fix change events

davidken91:


packages/js-example-app/src/index.ejs
```diff
@@ -133,6 +133,19 @@
             <p label="Tab 11" icon="blr360">Tab 11</p>
           </blr-tab-bar>
         </div>
+
+        <div class="component">
+          <p>Toggle Switch</p>
+          <blr-label-toggleswitch
+            theme="Light"
+            size="md"
+            showstatelabel="true"
```
angsherpa456, line 142: should be kebab-case.

packages/js-example-app/src/index.ejs
```diff
@@ -133,6 +133,19 @@
...
+            label="Label-text"
+            checkinputid="toggle-switchId"
```
angsherpa456, line 144: should be kebab-case.

packages/js-example-app/src/index.ejs
davidken91: Ok, I think that the kebab-case changes that you made a few weeks ago didn't take effect on the code snippet part of storybook. That's where I lifted this code from. We might need to fix that at a later time

packages/js-example-app/src/index.js
```diff
@@ -110,10 +113,46 @@ blrTextArea.addEventListener('blrBlur', () => {
...
+blrRadioInput.addEventListener('blrSelectedValueChange', (e) => {
+  addLog('blr-radio changed: ' + e.detail.selectedValue);
+});
+
+blrToggleSwitch.addEventListener('blrFocus', () => {
+  addLog('blr-label-toggleswitch focused');
+});
+
+blrToggleSwitch.addEventListener('blrBlur', () => {
+  addLog('blr-label-toggleswitch blurred');
```
angsherpa456, line 141: Can't it be just blr-toggleswitch blurred for the consistency of the naming with others. Same for line 145.

packages/ui-library/src/components/input-field-text/index.ts
```diff
@@ -89,8 +93,9 @@ export class BlrInputFieldText extends LitElementCustom {
   };
 
   protected handleChange = (event: Event) => {
+    const inputValue = this._inputFieldTextNode.value;
     if (!this.disabled) {
-      this.dispatchEvent(createBlrTextValueChangeEvent({ originalEvent: event }));
+      this.dispatchEvent(createBlrTextValueChangeEvent({ originalEvent: event, inputValue }));
```
angsherpa456, line 98: Can it be like ({ originalEvent: event, inputValue: this._inputFieldTextNode.value }) so that every single input does not have to create a inputValue variable also it is not used anywhere else.

packages/ui-library/src/components/radio/index.ts
```diff
@@ -36,6 +40,23 @@ export class BlrRadio extends LitElementCustom {
...
+
+  protected handleBlur = (event: FocusEvent) => {
+    if (!this.disabled) {
+      this.dispatchEvent(createBlrBlurEvent({ originalEvent: event }));
+    }
+  };
+
+  protected handleChange(event: Event) {
+    const selectedValue = this._radioNode.value;
+    this.dispatchEvent(createBlrSelectedValueChangeEvent({ originalEvent: event, selectedValue }));
```
angsherpa456, line 57: Same as above in packages/ui-library/src/components/input-field-text/index.ts

packages/ui-library/src/components/select/index.ts
```diff
@@ -88,7 +91,8 @@ export class BlrSelect extends LitElementCustom {
   }
 
   protected handleChange(event: Event) {
-    this.dispatchEvent(createBlrSelectedValueChangeEvent({ originalEvent: event }));
+    const selectedValue = this._selectNode.value;
+    this.dispatchEvent(createBlrSelectedValueChangeEvent({ originalEvent: event, selectedValue }));
```
angsherpa456, line 95: Same as above in packages/ui-library/src/components/input-field-text/index.ts

packages/ui-library/src/components/tab-bar/index.ts
```diff
@@ -172,14 +183,24 @@ export class BlrTabBar extends LitElementCustom {
                         role="tab"
                         aria-controls=${`panel-${index}`}
                         class="${navListItemClasses}"
-                        @click=${() => {
+                        @focus=${(event: FocusEvent) => {
+                          if (!isDisabled) {
```
angsherpa456, line 187: Please for consistency purpose move isDisabled inside the respective function and in the template just pass the handler like in our components. i.e. @focus={handleFocus}. Same for blur and click event.

packages/ui-library/src/components/textarea/index.ts
```diff
@@ -124,8 +127,9 @@ export class BlrTextarea extends LitElementCustom {
   }
 
   protected handleChange = (event: Event) => {
+    const inputValue = this._textareaNode.value;
```
angsherpa456, line 130: Same as above in packages/ui-library/src/components/input-field-text/index.ts

packages/ui-library/src/components/toggle-switch/index.ts
```diff
@@ -59,6 +60,12 @@ export class BlrToggleSwitch extends LitElementCustom {
     if (!this.disabled && !this.readonly) {
       this.onChange?.(event);
```
angsherpa456, line 61: Currently this piece of code is unique to toggle-switch only. My understanding is that if onChange prop is passed with the custom handler then execute that function rather than what we do in other components where we dispatch event like this 

`this.dispatchEvent(
        createBlrCheckedChangeEvent({
          originalEvent: event,
          checkedState: this.currentCheckedState,
        })
      );`
      
I don't know why it was implemented like this but here first checking if user had passed the prop or not might make sense. Eg:

if (this.onChange) {
this.onChange?.(event)
} else {
this.dispatchEvent(
        createBlrCheckedChangeEvent({
          originalEvent: event,
          checkedState: this.currentCheckedState,
        })
      );
}

Otherwise, two event will be dispatched while user might be only interested in one thing.
Same for blur and focus.

packages/ui-library/src/components/tab-bar/index.ts
davidken91: We can do that, but it won't match the consistency of other components because `isDisabled` is on a per tab basis instead of it being a component wide variable. We also need to have a check on the `keydown` event to make sure that the 'space' key was pressed. I've moved the disabled check to the function regardless, it might be a bit tidier

davidken91:
Close #1024

packages/ui-library/src/components/toggle-switch/index.ts
davidken91: I just realised that onChange shouldn't exist here. I'll update it to use blrChange now. It will work very similarly to checkbox

angsherpa456:
Changes looks good. Please adjust the number of commits.

davidken91:
> Changes looks good. Please adjust the number of commits.



> Changes looks good. Please adjust the number of commits.

Ok no problem. Do you mean that I should squash before I merge?

angsherpa456:
> > Changes looks good. Please adjust the number of commits.
> 
> Ok no problem. Do you mean that I should squash before I merge?

Yes, would be nice to have a single commit there.

angsherpa456:
> Ok, I think I can do this right before merging though github instead of the command line. Once I have two approvals, I should be able to complete this

With the regular workflow, would have been easier that way but after our develop diverted, would be easier if this is adjusted manually. Also currently "This branch is out-of-date with the base branch" state is there.

thrbnhrtmnn:
@davidken91 , cant we just merge this one? It was already approved and it would then close #1024

thrbnhrtmnn:
Ahh nvm, I see that you closed this and we merged the changes with #1066  right?

davidken91:
> Ahh nvm, I see that you closed this and we merged the changes with #1066 right?

Exactly. I updated the change events and the js example app in parallel without realising they were separate tickets

PR #1065, opened 2024-04-02
Fix: Text Area - Counter gets left-aligned if hasError=true #1064 

### LINKED ISSUE #1064: Text Area - Counter gets left-aligned if hasError=true
### What is affected?

Component

### Description

When I open the Text Area component in Storybook and set "hasError" to true (with default values), the counter gets left aligned but I expect the counter to still be right aligned.

Please also add a test to make sure the counter is always placed as expected, we had similar issues with counter placement before.

### Steps to reproduce

1. [Open Text Area component in storybook](https://b01ler.onrender.com/?path=/story/components-text-area--text-area)
2. Set "hasError" to true (leave everything else as defuault (hasCounter=true, errorMessage=empty, errorMessageIcon=undefined)
3. See that the Counter switches from being right aligned to being left aligned

### Affected version(s)

current develop version

Chrome Browser

### Code of Conduct

G-nizam-A:
Fixes #1064 

image

### DIFF
--- a/packages/ui-library/src/components/textarea/index.css.ts
+++ b/packages/ui-library/src/components/textarea/index.css.ts
@@ -44,9 +44,11 @@ export const styleCustom = typeSafeNestedCss`
 
     &.hint, &.error {
       justify-content: right;
+    }
 
-    &.sm      justify-content: space-between;
-    margin: ${captionslot.margin.sm};
+    &.sm {
+      & > blr-counter {
+        margin: ${captionslot.margin.sm};
       }
     }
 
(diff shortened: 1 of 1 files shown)


packages/ui-library/src/components/textarea/index.css.ts
```diff
@@ -43,7 +43,7 @@ export const styleCustom = typeSafeNestedCss`
     justify-content: right;
 
     &.hint, &.error {
-      justify-content: space-between;
+      justify-content: right;
```
davidken91, line 46: Thanks for your contribution 😊 This works for the error message, but I noticed that everything gets aligned right when we have both the counter and an error / hint. I was thinking that maybe we could use the :only-child pseudo class to determine if only the counter is there. That way we could use something like this:
`:only-child {
        margin-left: auto;
}
`

thrbnhrtmnn:
We decided today to close the PR, as the open review points have not been resolved and the already invested work is not sufficient to continue the bugfix on our own.

PR #1101, opened 2024-05-14
fix(ui-library): input field text ui fixes

### LINKED ISSUE #936: Text Input - color fixes
### Description / User story

In design the icon in the Text Input has a different color (darker) than in code. We should change it in code, so we are using the correct token.

- none

### Acceptance Criteria

- [ ] Color of icon in Text Input uses the correct tokens as defined in design

### Additional information

- ...

### Code of Conduct

ashk1996:
Closes #936

### DIFF
--- a/packages/ui-library/src/components/input-field-text/index.css.ts
+++ b/packages/ui-library/src/components/input-field-text/index.css.ts
@@ -86,10 +86,11 @@ export const styleCustom = typeSafeNestedCss/*css*/ `
             }
           }
 
-          &.disabled {
+          &[disabled] {
             &::placeholder {
               color: ${inputfield.placeholder.textcolor.default.disabled};
             }
+            color: ${inputfield.placeholder.textcolor.default.disabled};
           }
 
           &.focus {
@@ -166,6 +167,7 @@ export const styleCustom = typeSafeNestedCss/*css*/ `
         display: flex;
         justify-content: space-between;
         align-items: center;
+        background-color: ${inputfield.container.bgcolor.default.rest};
         border: ${inputfield.container.border.default.rest.width} ${inputfield.container.border.default.rest.style}
           ${inputfield.container.border.default.rest.color};
         border-radius: ${inputfield.container.borderradius};
@@ -203,7 +205,7 @@ export const styleCustom = typeSafeNestedCss/*css*/ `
           background-color: ${inputfield.container.bgcolor.default.focus};
         }
 
-        &.disabled {
+        &.disabled { 
           border-width: ${inputfield.container.border.default.readonly.width};
           border-style: ${inputfield.container.border.default.disabled.style};
           border-color: transparent;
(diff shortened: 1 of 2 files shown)


packages/ui-library/src/components/input-field-text/index.css.ts
```diff
@@ -86,10 +86,11 @@ export const styleCustom = typeSafeNestedCss/*css*/ `
             }
           }
 
-          &.disabled {
+          &[disabled] {
```
faselbaum, line 89: sorry for my earlier advice but i think we should change this back to a class selector since other styles seem to depend on this as well. see line 206. there also seems to be a color definition for the blr-input-wrapper in that line.

packages/ui-library/src/components/input-field-text/index.css.ts
```diff
@@ -86,10 +86,11 @@ export const styleCustom = typeSafeNestedCss/*css*/ `
...
             &::placeholder {
               color: ${inputfield.placeholder.textcolor.default.disabled};
             }
+            color: ${inputfield.placeholder.textcolor.default.disabled};
```
faselbaum, line 93: we should probably remove this. see previous comment

packages/ui-library/src/components/input-field-text/index.ts
```diff
@@ -218,7 +219,7 @@ export class BlrInputFieldText extends LitElementCustom {
                 </div>
               `
             : nothing}
-          <div class="blr-input-wrapper ${inputContainerClasses}">
+          <div class="blr-input-wrapper ${inputContainerClasses}" ?readonly="${this.readonly}">
```
faselbaum, line 222: in accordance with my first comment we should remove this attribute and add the disabled class to the `inputContainerClasses` object instead.

packages/ui-library/src/components/input-field-text/index.css.ts
ashk1996: Hi @faselbaum 
Thanks for pointing out. I have added the input color for disabled state within .blr-input-wrapper .blr-form-input
and removed  it here

packages/ui-library/src/components/input-field-text/index.css.ts
ashk1996: I have reverted this thanks for pointing out.

packages/ui-library/src/components/input-field-text/index.ts
ashk1996: I don't think we would need to remove this as disabled is already being applied to this class. The read only attribute is not being applied to this class hence i will keep it.

packages/ui-library/src/components/input-field-text/index.ts
```diff
@@ -170,6 +170,7 @@ export class BlrInputFieldText extends LitElementCustom {
       });
 
       const inputClasses = classMap({
+        'error-input': this.hasError || false,
```
faselbaum, line 173: this is also missing the `disabled` class. currently the placeholder has the wrong color in disabled state when compared to design.

ashk1996: resolved in the latest commit. Thank you

faselbaum (approved):

PR #1105, opened 2024-05-21
fix(ui-library): align counter to right for text area with or without error

### LINKED ISSUE #1064: Text Area - Counter gets left-aligned if hasError=true

ashk1996:
closes #1064

### DIFF
--- a/packages/ui-library/src/components/textarea/index.css.ts
+++ b/packages/ui-library/src/components/textarea/index.css.ts
@@ -38,9 +38,13 @@ export const staticStyles = typeSafeNestedCss/*css*/ `
       justify-content: right;
     }
     &.hint{
-      justify-content: space-between;
+      justify-content: right;
     }
     &.error-message
+    {
+      justify-content: space-between;
+    }
+    &.hint-message
     {
       justify-content: space-between;
     }
--- a/packages/ui-library/src/components/textarea/index.css.ts
+++ b/packages/ui-library/src/components/textarea/index.css.ts
@@ -34,10 +34,15 @@ export const staticStyles = typeSafeNestedCss/*css*/ `
   .blr-textarea-info-container {
     display: flex;
     justify-content: right;
-    &.hint {
-      justify-content: space-between;
-    }
   }
+  
+  .blr-textarea-info-container.error {
+    justify-content: right;
+  }
+  
+  .blr-textarea-info-container.hint {
+    justify-content: space-between;
+  }  
 
   .textarea-input-control {
     resize: none;
--- a/packages/ui-library/src/components/textarea/index.css.ts
+++ b/packages/ui-library/src/components/textarea/index.css.ts
@@ -40,6 +40,10 @@ export const staticStyles = typeSafeNestedCss/*css*/ `
     &.hint{
       justify-content: space-between;
     }
+    &.error-message
+    {
+      justify-content: space-between;
+    }
   }
 
   .textarea-input-control {
--- a/packages/ui-library/src/components/textarea/index.css.ts
+++ b/packages/ui-library/src/components/textarea/index.css.ts
@@ -34,10 +34,6 @@ export const staticStyles = typeSafeNestedCss/*css*/ `
   .blr-textarea-info-container {
     display: flex;
     justify-content: right;
-
-    &.error {
-      justify-content: right;
-    }
     &.hint {
       justify-content: space-between;
     }
--- a/packages/ui-library/src/components/textarea/index.css.ts
+++ b/packages/ui-library/src/components/textarea/index.css.ts
@@ -35,7 +35,10 @@ export const staticStyles = typeSafeNestedCss/*css*/ `
     display: flex;
     justify-content: right;
 
-    &.hint, &.error {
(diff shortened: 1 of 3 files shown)


packages/ui-library/src/components/textarea/index.css.ts
```diff
@@ -34,11 +34,15 @@ export const staticStyles = typeSafeNestedCss/*css*/ `
   .blr-textarea-info-container {
     display: flex;
     justify-content: right;
-
-    &.hint, &.error {
-      justify-content: space-between;
-    }
   }
+  
+  .blr-textarea-info-container.error {
```
angsherpa456, line 39: For consistency, please follow the similar pattern like other:
.blr-textarea-info-container {
&.error {...}
&.hit {...}
}

packages/ui-library/src/components/textarea/index.css.ts
```diff
@@ -34,11 +34,13 @@ export const staticStyles = typeSafeNestedCss/*css*/ `
...
-
-    &.hint, &.error {
+    &.error{
+      justify-content: right;
+    }
+    &.hint{
       justify-content: space-between;
     }
-  }
+  }
```
angsherpa456, line 43: Seems like your branch is out of sync with develop branch, please fix that and will approve it. Also please remove this extra space at line 43.

angsherpa456 (approved):
LGTM

thrbnhrtmnn (changes requested):
I just had a quick look and found one issue. If you add an error message and error message icon, they are right aligned, but they should be left aligned. 
Screenshot 2024-05-23 at 09 20 46

Here is how it should look according to design: 
Screenshot 2024-05-23 at 09 24 36

ashk1996:
> I just had a quick look and found one issue. If you add an error message and error message icon, they are right aligned, but they should be left aligned. Screenshot 2024-05-23 at 09 20 46
> 
> Here is how it should look according to design: Screenshot 2024-05-23 at 09 24 36

Hi @thrbnhrtmnn 
Just doubled checked, I see that it is left aligned. 
Screenshot 2024-05-23 at 10 08 13

thrbnhrtmnn:
> > I just had a quick look and found one issue. If you add an error message and error message icon, they are right aligned, but they should be left aligned. Screenshot 2024-05-23 at 09 20 46
> > Here is how it should look according to design: Screenshot 2024-05-23 at 09 24 36
> 
> Hi @thrbnhrtmnn Just doubled checked, I see that it is left aligned. Screenshot 2024-05-23 at 10 08 13

Hey @ashk1996 , in your example hasHint = true. In this case it looks good. But if hasHint = false and only hasError = true and errorMessage is not empty, then the error message is not left aligned but right aligned, as seen in my screenshot above. 

While looking at it again, if errorMessageIcon is not undefined and errorMessage is empty, the icon should also be displayed right aligned, but currently it is not displayed at all in this case. It will only appear once errorMessage is not empty anymore.

If you want we can have a quick huddle about this :-)

ashk1996:
> > > I just had a quick look and found one issue. If you add an error message and error message icon, they are right aligned, but they should be left aligned. Screenshot 2024-05-23 at 09 20 46
> > > Here is how it should look according to design: Screenshot 2024-05-23 at 09 24 36
> > 
> > 
> > Hi @thrbnhrtmnn Just doubled checked, I see that it is left aligned. Screenshot 2024-05-23 at 10 08 13
> 
> Hey @ashk1996 , in your example hasHint = true. In this case it looks good. But if hasHint = false and only hasError = true and errorMessage is not empty, then the error message is not left aligned but right aligned, as seen in my screenshot above.
> 
> While looking at it again, if errorMessageIcon is not undefined and errorMessage is empty, the icon should also be displayed right aligned, but currently it is not displayed at all in this case. It will only appear once errorMessage is not empty anymore.
> 
> If you want we can have a quick huddle about this :-)

Thanks for pointing out. I have fixed this in the latest commit.

thrbnhrtmnn:
Hey @ashk1996 , as discussed today the icon not being displayed if the corresponding message is empty is not an issue, because this would just be a user mistake. There should not be an icon when there is no message. Neither for errors nor for hints.

After looking again the only point I found is that in case the hasHint = true, but message and icon are empty / undefined, the counter will be left aligned. As just discussed we should fix this, as long it is not a huge effort. Again this would be an user mistake, but if we can prevent it easily we should always do so.

thrbnhrtmnn (approved):

PR #1115, opened 2024-06-07
fix(ui-library): using willupdate to manage property changes

### LINKED ISSUE #1083: Code quality: Improve consistency for property initialisation and defaulting
### Description / User story

As a developer, 
I want to have consistent initialisation and defaulting of property declarations so that we can get rid of inline nullish checking.
This is currently very inconsistent across component implementations and bloats our code.

Eg.: Some components declare and initialize properties like this

```typescript
@property sizeVariant?: FormSizesType = 'md'
```

VS

```typescript
@property sizeVariant: FormSizesType = 'md'
```

The latter one means that from a compiler perspective `sizeVariant` will always be defined and does not need to be validated upon usage, reducing nullish checking operations in many parts of the code. This leads to better readability. This is just a simple example though and doesn't account for the fact that our properties and state are potentially initialized from attribute values or might be changed to invalid values from javascript.

We need to find a general strategy on how exactly we want to do this and then implement this strategy for all components.

### Possible Solutions

A) Use custom accessors for defaulting and sanitizing - https://lit.dev/docs/components/properties/#accessors
B) Use `willUpdate` lifecycle callback for defaulting and sanitizing - https://lit.dev/docs/components/lifecycle/#willupdate

- none

### Acceptance Criteria

- [ ] If an attribute is set to an invalid value we need to default it to a valid one for each component (maybe start with only a few components)
- [ ] All of our properties (in at least 5 components) should not be defined with a ? operator, except where undefined is a valid value
- [ ] At least 5 components have been updated to the new approach
- [ ] If not all components where updated with this task, a follow-up task has been created

### Additional information
- ...

### Code of Conduct

ashk1996:
closes #1083

### DIFF
--- a/packages/ui-library/src/components/radio-group/index.test.ts
+++ b/packages/ui-library/src/components/radio-group/index.test.ts
@@ -5,7 +5,6 @@ import { BlrRadioGroupRenderFunction } from './renderFunction';
 import { fixture, expect } from '@open-wc/testing';
 import { querySelectorDeep } from 'query-selector-shadow-dom';
 import { BlrRadioGroupType } from '.';
-import { PropertyValueMap } from 'lit';
 
 const sampleParams: BlrRadioGroupType = {
   theme: 'Light',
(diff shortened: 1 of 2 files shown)


ashk1996:
Closed pr as we have another active PR https://github.com/deven-org/boiler/pull/1127

PR #1123, opened 2024-06-13
fix(storybook): checkbox size issue

### LINKED ISSUE #754: Checkbox - fix size issues
#### Description / User Story

The Checkbox control width gets smaller, when the label gets too long.

See:
checkbox-bug

----
#### Requirements
- none

----
#### Acceptance Criteria
- [ ] The Checkbox control does not change in size, regardless how long the label is.
- [ ] If possible: Tests have been added to the Checkbox to prevent these issues from happening again

----
#### Background information
- ...

bar-tay:
closes #754

### DIFF
--- a/packages/ui-library/src/components/checkbox/index.css.ts
+++ b/packages/ui-library/src/components/checkbox/index.css.ts
@@ -47,7 +47,7 @@ export const staticStyles = typeSafeNestedCss/* css */ `
           gap: ${checkbox.contentrow.itemspacing.sm};
           
           .visual-checkbox {
-            width: ${checkbox.control.container.size.sm};
+            min-width: ${checkbox.control.container.size.sm};
             height: ${checkbox.control.container.size.sm};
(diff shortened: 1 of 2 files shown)


thrbnhrtmnn (approved):
Hey @bar-tay , this looks good to me in Storybook. One thing, would it be possible to implement a test, which checks that the width of the checkbox is always as desired without mich effort?

bar-tay:
@thrbnhrtmnn should this be an accessibility test which is shown in the tab storybook or a regular test in the code?

faselbaum (approved):

thrbnhrtmnn:
> @thrbnhrtmnn should this be an accessibility test which is shown in the tab storybook or a regular test in the code?

No, this is not about a11y, it would be a regular test in code to decrease the chances of this bug happening again in the future.

bar-tay:
@thrbnhrtmnn I don't think that's necessary, since I changed the width of the checkbox with a property which ensures that it   can't be any other size than the desired one.

PR #1133, opened 2024-07-02
feat(storybook): select icon clickable

bar-tay:


packages/ui-library/src/components/select/index.ts
```diff
@@ -108,7 +115,7 @@ export class BlrSelect extends LitElementCustom {
         this.sizeVariant,
       ]).toLowerCase() as SizesType;
 
-      const modifiedIcon = this.icon ? this.icon : 'blrChevronDown';
+      const modifiedIcon = this.icon;
```
angsherpa456, line 118: This is used in only one place and also you can directly use the this.icon so this variable can be removed.

packages/ui-library/src/components/select/index.ts
```diff
@@ -119,9 +126,10 @@ export class BlrSelect extends LitElementCustom {
         },
         {
           'aria-hidden': true,
-        }
+        },
       );
     }
+    return nothing;
```
angsherpa456, line 132: Is this return necessary?

thrbnhrtmnn:
Hey @bar-tay ,

here are my findings from the review:
- The icon is still shown, even if icon = undefined
- Would it be possible to increase the clickable area of the select so a user can click everywhere inside the outlined area and the menu opens?

packages/ui-library/src/components/select/index.ts
bar-tay: done

packages/ui-library/src/components/select/index.ts
bar-tay: done

packages/ui-library/src/components/select/index.ts
```diff
@@ -176,37 +184,39 @@ export class BlrSelect extends LitElementCustom {
...
+                ${this.renderIcon()}
+              </div>
             </div>
-            ${this.renderIcon()}
           </div>
           ${(this.hasHint && this.hintMessage) || (this.hasError && this.errorMessage)
             ? BlrFormCaptionGroupRenderFunction(
                 { theme: this.theme, sizeVariant: this.sizeVariant },
-                this.renderCaptionContent()
+                this.renderCaptionContent(),
```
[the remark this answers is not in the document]
bar-tay: done

thrbnhrtmnn:
Hey @bar-tay ,
thanks for the update, this looks good now.

As discussed, lets change 2 minor things:
1. change the default option to the first in the array
2. change the storybook setting, so the parent container is the same as the one we use for input field text & number (using the whole width of the viewport)

ashk1996 (approved):

thrbnhrtmnn (approved):

PR #1136, opened 2024-07-04
fix(ui-library): 1135 tab component slot

### LINKED ISSUE #1135: TabBar - Create a TabBarItem component and pass it as a slotted elements
### Description

Currently the TabBar component gets the paragraph elements as the children which is not so optimal. Inside the TabBar component the use of slot is not as it should be. In this ticket, the current approach should be adjusted by creating a new TabBarItem component and use that instead of paragraph element as the slotted element and inside the TabBar component the proper use of slot is achieved.

### Code of Conduct

angsherpa456:
closes #1135

### DIFF
--- a/packages/js-example-app/src/index.client.ts
+++ b/packages/js-example-app/src/index.client.ts
@@ -203,14 +203,6 @@ function init() {
     addLog('blr-toggleswitch changed: ' + e.detail.checkedState);
   });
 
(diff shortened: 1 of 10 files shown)


thrbnhrtmnn (approved):
Tab Bar in Storybook looks as expected

bar-tay (approved):

PR #1137, opened 2024-07-11
feat(storybook): use outline instead of border

### LINKED ISSUE #1109: Text Area, Input Field Text, Input Field Number, Select, Counter - use outline instead of border consistently with right offset
### Description / User story

Text Area, Input Field Text, Input Field Number, Select and Counter should use outlines instead of borders and the outlines should also be offset, similar to how its done in Button Text (secondary variant) or in the unchecked Checkbox component.

The reason for this was elaborated within #602 and it has to do with how we are displaying borders in Figma. In Figma borders are not changing the size of a container and they are always applied inside the container. In code a border of 1px will increase the total size of a container by 1px in all directions where the border is applied. To be aligned with code we decided to add an offset, which is exactly the same as the border/outline width * -1 (see Button Text secondary variant for example).

Figma has since added the option to include borders in the size of a container. A switch on design side however would also probably cause flickering behaviour for when we are switching the border from 1px to 2px in focus states. Therefor we decided to stick with the offset solution.

The goal of this task is to align all components that have borders/outlines and implement the offset solution as well as the usage of borders & outlines consistently.

- [x] none

### Acceptance Criteria

- [ ] Text Area uses the same approach to outlines and borders than we use in the Button Text secondary variant or in the unchecked Checkbox components.
- [ ] Input Field Text uses the same approach to outlines and borders than we use in the Button Text secondary variant or in the unchecked Checkbox components.
- [ ] Input Field Number uses the same approach to outlines and borders than we use in the Button Text secondary variant or in the unchecked Checkbox components.
- [x] Counter uses the same approach to outlines and borders than we use in the Button Text secondary variant or in the unchecked Checkbox components.
- [x] Select uses the same approach to outlines and borders than we use in the Button Text secondary variant or in the unchecked Checkbox components.

### Additional information
- This task is related to #1118

### Code of Conduct

bar-tay:
closes #1109

### DIFF
--- a/docs/GETSTARTED.md
+++ b/docs/GETSTARTED.md
@@ -10,7 +10,10 @@ seasoned professional or a newcomer, we've got you covered. Let's dive in and un
     - [Designers](#designers)
     - [Developlers](#developers)
     - [Product & Project Management](#product--project-management)
-  - [Regular Meetings](#regular-meetings)
+  - [Meetings](#meetings)
+    - [Regular Meetings](#regular-meetings)
+    - [Optional Meetings](#optional-meetings)
+    - [Paused Meetings](#paused-meetings)
   - [Communication Channels](#communication-channels)
   - [Access Checklist](#access-checklist)
   - [Project Management Onboarding Checklist](#project-management-onboarding-checklist)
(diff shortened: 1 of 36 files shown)


thrbnhrtmnn (changes requested):
Hey @bar-tay ,
all components mentioned in the tickets are using an outline now, but I can not find the calculation to offset the outline in any of the components when I check them in Storybook. Can you check again?

ashk1996 (approved):

thrbnhrtmnn (commented):
Hey @bar-tay ,

the changes all look as desired, but I think you forgot to do the same for the Counter component. Also it would be really nice, if you could also add the outline offset calc to the focus state for all three input components (text area, input field text, input field number).

thrbnhrtmnn:
Hey @bar-tay , I also just realized, that I forgot to add the Select component to the ACs. This one should also have the same outline offset, also for the focus state if possible.

bar-tay:
@thrbnhrtmnn 
I adjusted it

thrbnhrtmnn (changes requested):
Hey @bar-tay ,

thank you for the adjustment! All components have the offset in their default states now. However it seems to me, that you used hard values to define the offset. This is not how it is intended. Instead the calculation should use the token for the outline-width and multiply it with -1 to make it negative.

Another open point is the focus states. For the select this loos great now but for Text Area, Input Field Text and Input Field Number the Focus state did not change.

bar-tay:
@thrbnhrtmnn 
I changed it

thrbnhrtmnn (commented):
Hey @bar-tay ,
almost perfect now. We are only missing the outline offset for the focus state in the text area component now. After that its done.

bar-tay:
@thrbnhrtmnn done

thrbnhrtmnn (approved):
Looks good in Storybook

Issue #933, opened 2024-02-19
Input Field Number - fix input background, value and unit color


thrbnhrtmnn:
### Description / User story

In design the input field of the Input Field Number has a non-transparent background color. In code the color is transparent and should be changed to the correct color, defined by design.

~Furthermore the unit color is different when the value is empty and when the value is filled in design, but in code it is always the same. The unit color for when the value is empty needs to be added in code.~ It should be the same, as it now is also in design.

Another missing color when comparing design and code is the unit, placeholder and value in the error state. According to design the value color is a dark red in the error state, in code the color is the same as in other states (dark-grey/black/white).

### Requirements / Prerequisites

- none

### Acceptance Criteria

- [x] The background color of the input field in the Input Field Number in code is the same as in design in light and dark mode (also in regards to transparency)
- [x] The stepper icon color of the input field in the Input Field Number in code is the same as in design in light and dark mode in disabled and readonly state (also in regards to transparency)
- [x] The color of the value in the Input Field Number in code is the same as in design in error state. (might also be fixed with #683, in case we are already using the right token)
- [ ] The color of the unit in the Input Field Number in code is the same as in design in error state. (might also be fixed with #683, in case we are already using the right token)
- [x] The color of the placeholder in the Input Field Number in code is the same as in design in error state. (might also be fixed with #683, in case we are already using the right token)

### Additional information

- A related task is #831 , where the properties "unit" and "prependUnit" / "hasUnit" are renamed/changed.
- Font color of error state is also part of #683 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #932, opened 2024-02-19
Input Field Number - fix width of horizontal stepper variant


thrbnhrtmnn:
### Description / User story

The width of the stepper buttons in the stepper variant "horizontal" of the Input Field Number is not aligned between design and code.

In design the each stepper button should be squared (i.e. 32x32 in SM), but in code the width is smaller than it should be and even changes when the width of the wrapping container of the Input Field Number changes.

The width in code should be aligned with the width in design.

### Requirements / Prerequisites

- none

### Acceptance Criteria

- [x] The width of the Stepper Buttons in the "horizontal" stepper variant of the Input Field Number is the same in code as in design (32 px in SM, 40 px in MD and 48 px in LG)

### Additional information

- ...

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

AnukratiMehta:
I'd like to give this a try @thrbnhrtmnn.

thrbnhrtmnn:
Hey @AnukratiMehta , thanks for reaching out and for wanting to contribute to the project!

Unfortunately we are currently doing a massive update to our CSS, so I would highly recommend to wait a bit before starting this issue. We probably will finish it this week, maybe even today. I will let you know when this can be started :-)

thrbnhrtmnn:
Hello @AnukratiMehta , it took a bit longer, but we just finished the CSS update. If you still want to give this a try, please go ahead :-)

Issue #915, opened 2024-02-14
Input Field Number - align readonly & disabled state with design


thrbnhrtmnn:
### Description / User story

The readonly state of the Input Field Number should not include the stepper buttons according to design.
We had this status in the past already, but accidentally added the stepper buttons again to the readonly state with https://github.com/deven-org/B01LER-Kitchen/pull/719 .

Also in disabled state the color of the stepper button icons should be grey instead of blue.

### Requirements / Prerequisites

- none

### Acceptance Criteria

- [x] When in readonly state, the Input Field Number does not display any stepper buttons and the value can not be adjusted.
- [x] When in disabled state, the stepper buttons of the Input Field Number have the same color as defined in design (in all stepper variants) - While fixing this, the stepper buttons should also be checked in other states in dark mode, as they have changed in design and are now #fff
- [x] When in disabled state, the input/value color has the same color as defined in design

### Additional information

- ...

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1152, opened 2024-09-02
fix(storybook):  state issues

bar-tay:
closes


thrbnhrtmnn (changes requested):
Hey @bar-tay , here are my findings:

#933
- The unit color is not correct in disabled and in error state

#932
- The widths are all correct now, but it seems like the height in the MD is wrong now for the vertical stepper variant

#915
- When a unit is shown or when stepperVaraint = Split, the readonly state now shows some outline or border around the user input container

thrbnhrtmnn (approved):
Thanks @bar-tay , everything has been solved :-)

PR #1161, opened 2024-09-26
fix(ui): a11y test

### LINKED ISSUE #778: ToggleSwitch - fix accessibility violations
#### Description / User Story

The following errors and incomplete tests are thrown by the storybook accessibility addon and should be fixed:
- Incomplete: Form field must not have multiple label elements (all states & variants)

----
#### Requirements
- [x] #542 

----
#### Acceptance Criteria
- [ ] All errors and incomplete tests thrown by the accessibility addon are resolved:
   - [ ] Incomplete: Form field must not have multiple label elements (all states & variants)

----
#### Background information
- ...

bar-tay:


thrbnhrtmnn (changes requested):
Hey @bar-tay ,
i still find a11y errors in the toggle switch when looking at the component. It also seems like we have more errors now than the one that is stated in the ticket. We should fix all errors with this PR:
- Violation: "Elements in the focus order should have an appropriate role", happens in all states
- Incomplete: "Form field must not have multiple label elements", happens when hasStateLabel=true and label=true
- Violation: "Form elements must have labels", happens if hasStateLabel=false and label=false; here I am not sure if this is okay as this variant of the component would require a text somewhere next to it which would need to function as the label

Furthermore I could not check the dark mode anymore because it broke with the PR. On develop the dark mode is still working.

ashk1996 (approved):

bar-tay:
The approach I attempted to implement did not work as expected. It would either break the component or generate new accessibility errors. The "Violation" error in the default and other states, which was apparently already present in the develop branch, persisted even when I applied the appropriate role to the component.
It seems to me that this component might need some restructuring to resolve all the accessibility errors.

PR #1162, opened 2024-09-30
feat(release): use release process on main

ashk1996:


ashk1996:
a new pr will be created. testing on the forked repository https://github.com/ashk1996/boiler/pull/1

PR #1167, opened 2024-10-02
fix(ui-library): removed readonly state for checkbox radio radiogroup and toggle switch

ashk1996:
Removed readonly state and css for the above components


thrbnhrtmnn (approved):
@ashk1996 , PR looks good to me, before merging, we also need to do the design changes. Please contact @larserbach for this

larserbach (approved):

PR #1172, opened 2024-11-06
fix(ui-library): pressed state removed and updated focus state for readonly

### LINKED ISSUE #1125: Text Area, Input Field Text, Input Field Number - align behaviour when focusing component in readonly state
### Description / User story

Currently the Text Area, Input Filed Text and Input Field Number components each behave a bit different when they are focused while in readonly state. This task includes the sync on this with design and the alignment of all 3 components.

- none

### Acceptance Criteria

- [x] The correct behaviour when focusing the 3 components in readonly state has been found together with design:
   - inputs are focusable in readonly state
   - input values must be selected if present (regardless of the state)
- [ ] The Text Area can be focused in readonly state and doing so will select the value
- [x] The Input Field Text can be focused in readonly state and doing so will select the value
- [x] The Input Field Number can be focused in readonly state and doing so will select the value

### Additional information

- ...

### Code of Conduct

### LINKED ISSUE #1155: Text Area, Input Field Text, Input Field Number, Select - remove pressed state & update focus state
### Description / User story

As researched and proposed in #1125 the pressed state should be removed for the Text Area, Input Field Text, Input Field Number and Select components, as it is not used for input fields.

Further more the focus state should be updated so it only contains the needed tokens for storke-color, stroke-width and potentially background-color. In code the focus state should also always dominate other states and overwrite all other tokens.

- none

### Acceptance Criteria

- [ ] The pressed state has been removed in **design** from:
   - [ ] Text Area
   - [ ] Input Field Text
   - [ ] Input Filed Number
   - [ ] Select
- [x] The pressed state has been removed in **code** from:
   - [x] Text Area
   - [x] Input Field Text
   - [x] Input Filed Number
   - [x] Select

- [x] The focus state of the following components in code uses all tokens as defined in design:
   - [x] Text Area
   - [x] Input Field Text
   - [x] Input Filed Number
   - [x] Select

- [x] The focus states dominates other states like hover or readonly for the following components in **code**:
   - [x] Text Area
   - [ ] Input Field Text
   - [x] Input Filed Number
   - [x] Select

### Additional information

- ...

### Code of Conduct

ashk1996:
closes #1155 
closes #1125


thrbnhrtmnn (changes requested):
Hey @ashk1996 ,
here are my findings:

#1155 :
- Could we update the input field text, so that the focus ring gets applied as soon as I press down the mouse button? Currently it will only get applied once I release the button (for all other components its working this way already)
- The Input Field hover state when hasError=true looks different than in design (both in dark and light mode) it should look the same as the rest state. Also the focus state is not dominating this state.

#1125 :
- Focusing a readonly Text Area component via a Tab / via keyboard navigation, does not select the value (for input field text and input field number this works).

thrbnhrtmnn (approved):
LGTM

@ashk1996 , maybe Christian could also have a quick look on Monday, for me it can be merged :-)

ChristianHoffmannS2 (approved):

PR #1204, opened 2025-01-07
feat(tokens): adds index-generator

larserbach:
PR adds script to generate "_tokens/generated/index.pseudo.generated.ts"


larserbach:
@ChristianHoffmannS2 To be discussed:
- How can we improve the name "index.pseudo.generated.ts"? If we remove the "pseudo" the filename becomes imho to generic.
- What could be a fitting name for the generator file?

ChristianHoffmannS2 (approved):

ChristianHoffmannS2:
> @ChristianHoffmannS2 To be discussed:
> 
> * How can we improve the name "index.pseudo.generated.ts"? If we remove the "pseudo" the filename becomes imho to generic.
> * What could be a fitting name for the generator file?

how about iterators.generated.ts?

ashk1996 (approved):

Issue #1156, opened 2024-09-11
Checkbox - remove gap to the left


thrbnhrtmnn:
### Description / User story

When comparing design and code of the Checkbox a few misalignments can be seen:
- In Code the checkbox seems to have a gap left to the control, which is not there in design. The gap to the left should be removed
- ~The focus ring in design and code seem to have different values, but they look the same. In sizeVariant=MD the container is 24x24px in design but 16x16px in code. Also the border-radius of the focus ring seems to not be defined by a token in design, but just is 5px, whereas in code we apparently use the token for the checkbox control to calculate the border radius of the focus ring. I am afraid this approach would not work anymore when the control should not have a radius but the focus ring should have one.~

### Requirements / Prerequisites

- none

### Acceptance Criteria

- [x] The checkbox does not have a gap left of the control on code
- [x] ~The focus ring in design is fully tokenized~
- [x] ~The focus ring in code uses all tokens that are defined in design to look the same as in design~

### Additional information

- ...

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

thrbnhrtmnn:
As decided today, the tokenizing of the focus layer in the checkbox component should be done separately and at a later point. This requires some design research.

PR #1232, opened 2025-02-09
fix(ui-library): checkbox component bug fixes

### LINKED ISSUE #1157: Checkbox - multiple behaviour fixes in code
### Description / User story

The checkbox currently seems to have some issues where it is not behaving as intended:
- Currently, when indeterminate icon gets clicked it gets checked. I believe this is wrong and the value should not change. This is how it is behaving in [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes).
- When hasError=true, the component seems to have an empty blr-form-caption-group container. This should not be there unless errorMessage or errorMessageIcon also have a value. Compare this to radio, where this works as intended.
- When hasHint=true, but hintMessage and hintMessageIcon are empty, the component seems to have an empty blr-form-caption-group container. This should not be there. Compare this to radio, where this works as intended.
- The checkbox currently switches to checked/unchecked, when the mouse button is pressed down. I believe the checking/unchecking should only happen once the mouse button is released, while the cursor is still within the control. This is how it behaves in [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox).
- When hasError=true, switching indeterminatedt to true does not show the indeterminated icon as defined in design. This needs to be fixed in code.

- none

### Acceptance Criteria

- [x] Clicking on the indeterminate checkbox does not change the value
- [x] When hasError=true, but errorMessage and errorMessageIcon are empty/undefined there the caption container does not contain an empty blr-form-caption-group
- [x] When hasHint=true, but hintMessage and hintMessageIcon are empty/undefined there the caption container does not contain an empty blr-form-caption-group
- [x] When clicking the checkbox control, the switch from unchecked to checked and vice-versa happens once the pressed down mouse button gets released while the cursor is within the control
- [x] When hasError=true and indeterminated=true, the checkbox shows the indeterminated checkbox in the error state.

### Additional information

- ...

### Code of Conduct

ashk1996:
closes #1157 #1156

### DIFF
--- a/packages/ui-library/src/components/checkbox/index.css.ts
+++ b/packages/ui-library/src/components/checkbox/index.css.ts
@@ -17,6 +17,7 @@ export const staticStyles = css`
 
         .control-wrapper {
           display: flex;
+
           .visual-checkbox {
             all: initial;
             display: inline-block;
--- a/packages/ui-library/src/components/checkbox/index.css.ts
+++ b/packages/ui-library/src/components/checkbox/index.css.ts
@@ -27,6 +27,9 @@ export const staticStyles = css`
 
             & .checker-icon {
               position: absolute;
+              display: flex;
+              align-items: center;
+              justify-content: center;
             }
           }
           .label-wrapper {
@@ -299,7 +302,8 @@ export const staticStyles = css`
           }
 
           &.error {
-            &.checked {
+            &.checked,
+            &.indeterminate {
               background-color: ${checkbox.control.container.bgcolor.active.error};
               outline-color: ${checkbox.control.container.bordercolor.active.error};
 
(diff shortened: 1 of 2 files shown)


ChristianHoffmannS2 (approved):

thrbnhrtmnn (changes requested):
Hey @ashk1996 ,

you completed all acceptance criteria so this new checkbox already looks really nice!

However I found one issue so far: The size of the icon so either the Checkmark or the Minus are way to big in SM and MD and might also be slightly off in LG. Could you check this again?

ashk1996:
Hi @thrbnhrtmnn 

The checkmark/minus icon issue was a bug I noticed, possibly caused by another fix. Although it wasn't within the scope of the ticket, I implemented a quick fix to ensure it displays correctly.

It would be great to create a separate ticket for this issue. Would that be okay?

thrbnhrtmnn (approved):
Hey @ashk1996 , okay yes lets move the icon fix into another issue. In that case I approve the task.

